PR #14, opened 2019-01-14
Improve type coverage

muglug:
This tightens up some of the types used in the project, ensuring things are never set to values they cannot be, and adding closure params/return types. It also avoids calling methods that give the same result more than once.


mglaman:
@muglug thanks! Being in Drupaland still stuck supporting PHP 5.6 until spring. So finally using typing is new to me.

PR #18, opened 2019-01-18
Remove typehint to allow all types of variables

### LINKED ISSUE #17: Typehint in DrupalExtension produces PHP Fatal error
Introduced in #14.

The added typehint leads to a fatal error when a parameter for a service definition is not a string (e.g. `null`, `array`).

https://github.com/mglaman/phpstan-drupal/blob/410ea5ad288455a206693ced0c890df80524c1f2/src/DependencyInjection/DrupalExtension.php#L145

I'd propose to remove the typehint until we come up with a better solution.

daFish:
Fixes #17


mglaman:
In #29 I made a fix which excludes the one case this happens.

PR #19, opened 2019-01-18
[WIP]: Added parameter support to specify docroot

### LINKED ISSUE #16: Enable configuration of docroot
Right now the extension assumes you have either `docroot` or `web` as your docroot. However, in one of our instances we use `public_html` and cannot chance this for the time being.

It would be great if we could configure the used docroot. If someone can point to the right extension points I would be more than interested to contribute.

daFish:
This change is Work in Progress.

- [x] add support to specify docroot in `phpstan.neon` under `parameters:`-key
- [ ] add fallback in case parameter was not specified explicitly 
- [ ] utilize parameter in `PHPStan\DependencyInjection\DrupalExtension`
- [ ] utilize parameter in `PHPStan\Drupal\Bootstrap`

Fixes #16

### DIFF
--- a/src/DependencyInjection/DrupalExtension.php
+++ b/src/DependencyInjection/DrupalExtension.php
@@ -55,6 +55,7 @@ class DrupalExtension extends CompilerExtension
 
     public function loadConfiguration(): void
     {
+        $builder = $this->getContainerBuilder();
 
         $this->autoloaderPath = $GLOBALS['autoloaderInWorkingDirectory'];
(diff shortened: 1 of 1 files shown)


daFish:
@mglaman I could need some help getting the configuration into the bootstrap process. Any thoughts on that?

jibran:
Let's use https://github.com/webflo/drupal-finder instead.

mglaman:
It seems like #25 did the trick, closing.

PR #26, opened 2019-02-02
[#24] Adds profiles namespaces.

### LINKED ISSUE #24: Add profile namespaces.
Right now if you have classes/tests in custom profile `\PHPStan\Drupal\Bootstrap::register()` doesn't add them to autoloader.

jibran:
Fixes #24.
Signed-off-by: Jibran Ijaz <jibran.ijaz@gmail.com>

### DIFF
--- a/src/Drupal/Bootstrap.php
+++ b/src/Drupal/Bootstrap.php
@@ -74,7 +74,7 @@ class Bootstrap
         }, $profiles);
         $this->extensionDiscovery->setProfileDirectories($profile_directories);
 
-        $this->moduleData = $this->extensionDiscovery->scan('module');
+        $this->moduleData = array_merge($this->extensionDiscovery->scan('module'), $profiles);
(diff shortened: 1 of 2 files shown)


src/Drupal/Bootstrap.php
```diff
@@ -74,7 +74,7 @@ public function register(): void
         }, $profiles);
         $this->extensionDiscovery->setProfileDirectories($profile_directories);
 
-        $this->moduleData = $this->extensionDiscovery->scan('module');
+        $this->moduleData = array_merge($this->extensionDiscovery->scan('module'), $profiles);
```
jibran: If we want to rename the variable or method name to `extensionData` or  `addExtensionNamespaces` feel free to do that before commit. I just wanted to show the fix here. :)

mglaman: I think this is fine. Technically themes are extensions as well. But profiles are more like "special" modules anyways.

PR #27, opened 2019-02-02
Entities: Allow FieldItemList access through magic __get/__set methods

Fonata:
PHPStan said "Access to an undefined property". A custom entity may
define fields in ::baseFieldDefinitions(). Other ways to define a
field are though hook implementations like hook_entity_base_field_info(),
or through config entites created from the Drupal admin UI. Thus,
it seems very hard to track all cases in PHPStan.

While ContentEntityBase::__get() will return a FieldItemListInterface,
ContentEntityBase::__set may be used to set fields directly, thus I
decided to tell PHPStan that the type is mixed. For example, the following
code can be valid:

````php
$entity = SomeCustomEntity();
$entity->some_int_field = 34;
$entity->save();
````

### CODE REVIEW

**Comment by Fonata** [comment]
@mglaman Thanks for your time and feedback. 

Should I go ahead and make Travis happy? I will if you agree with the general approach.

File: src/Reflection/EntityFieldsViaMagic.php [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Allows field access via magic methods
+ *
+ * See \Drupal\Core\Entity\ContentEntityBase::__get and ::__set.
+ */
+class EntityFieldsViaMagic implements PropertiesClassReflectionExtension {
+
+  public function hasProperty(ClassReflection $classReflection, string $propertyName): bool {
+    return $classReflection->isSubclassOf(ContentEntityBase::class) &&
```
mglaman: We should target the specific class which supports `__get` and `__set` to the `$values` and `$fields` properties. I believe this is `FieldableEntityInterface`.

src/Reflection/EntityFieldsViaMagic.php
```diff
@@ -0,0 +1,53 @@
...
+      !$classReflection->hasNativeProperty($propertyName);
+  }
+
+  public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection {
+    return new class implements PropertyReflection {
+
+      public function getType(): \PHPStan\Type\Type {
+        return new MixedType();
```
mglaman: It's not a mixed type, we should return `FieldItemList`. I think it is _safe enough_ to assume no custom field type is writing a FieldItemListInterface implementation which builds off of Core. Worth noting that, however.

src/Reflection/EntityFieldsViaMagic.php
Fonata: You're right, `FieldableEntityInterface` is the perfect fit. A lot of code I have seen incorrectly uses its parent interface, `EntityInterface`, though. Here is an example from Core:
````php
public function Drupal\Core\Entity\EntityListBuilder::buildRow(EntityInterface $entity);
````

Please decide if which interface we should ask for.

**Comment by Fonata** [comment]
I messed one of the commits up: I had no intention of changing `DrupalExtension.php`, so I added a fixup commit. Git will fix things for us when running `git rebase --autosquash`.

**Comment by Fonata** [comment]
@mglaman When you find the time: Please let me know whether we're moving in the right direction. Once I know the code here is stable enough, I will try to add PHPUnit test coverage. Thanks.

**Comment by Fonata** [comment]
I just rebased this to the current master branch as there were conflicts.

**Comment by mglaman** [comment]
@Fonata sorry I've been busy. I'm working to make time to come back over here.

**Comment by Fonata** [comment]
No problem at all, your pace is fine.

**Comment by mglaman** [comment]
:D We have test coverage! Let's jam on this.

**Comment by mglaman** [comment]
@Fonata sorry I screwed up the branch a bit, but the diff is OK enough for a squash merge. Working on the tests.

**Comment by mglaman** [comment]
With the test and without the broker addition the output should be:

```
Array
(
    [0] => PHPStan\Analyser\Error Object
        (
            [message:PHPStan\Analyser\Error:private] => Access to an undefined property Drupal\Core\TypedData\TypedDataInterface::$value.
            [file:PHPStan\Analyser\Error:private] => /Users/mglaman/Projects/php/phpstan-drupal/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityFieldFixture.php
            [line:PHPStan\Analyser\Error:private] => 18
            [canBeIgnored:PHPStan\Analyser\Error:private] => 1
        )

    [1] => PHPStan\Analyser\Error Object
        (
            [message:PHPStan\Analyser\Error:private] => Access to an undefined property Drupal\entity_test\Entity\EntityTest::$name.
            [file:PHPStan\Analyser\Error:private] => /Users/mglaman/Projects/php/phpstan-drupal/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityFieldFixture.php
            [line:PHPStan\Analyser\Error:private] => 19
            [canBeIgnored:PHPStan\Analyser\Error:private] => 1
        )

    [2] => PHPStan\Analyser\Error Object
        (
            [message:PHPStan\Analyser\Error:private] => Access to an undefined property Drupal\entity_test\Entity\EntityTest::$name.
            [file:PHPStan\Analyser\Error:private] => /Users/mglaman/Projects/php/phpstan-drupal/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityFieldFixture.php
            [line:PHPStan\Analyser\Error:private] => 20
            [canBeIgnored:PHPStan\Analyser\Error:private] => 1
        )

)
```

**Comment by mglaman** [comment]
With the extension added to the broker:

```
        $label1 = $testEntity->label(); // 17
        $label2 = $testEntity->get('name')->first()->value; // 18
        $label3 = $testEntity->name->first()->value; // 19
        $label4 = $testEntity->name->value; // 20
```

It fixed line 20! which is what we intended to do, here.

```
Array
(
    [0] => PHPStan\Analyser\Error Object
        (
            [message:PHPStan\Analyser\Error:private] => Access to an undefined property Drupal\Core\TypedData\TypedDataInterface::$value.
            [file:PHPStan\Analyser\Error:private] => /Users/mglaman/Projects/php/phpstan-drupal/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityFieldFixture.php
            [line:PHPStan\Analyser\Error:private] => 18
            [canBeIgnored:PHPStan\Analyser\Error:private] => 1
        )

    [1] => PHPStan\Analyser\Error Object
        (
            [message:PHPStan\Analyser\Error:private] => Access to an undefined property Drupal\Core\TypedData\TypedDataInterface::$value.
            [file:PHPStan\Analyser\Error:private] => /Users/mglaman/Projects/php/phpstan-drupal/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityFieldFixture.php
            [line:PHPStan\Analyser\Error:private] => 19
            [canBeIgnored:PHPStan\Analyser\Error:private] => 1
        )

)
```

File: src/Reflection/EntityFieldReflection.php [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $this->declaringClass = $declaringClass;
+        $this->propertyName = $propertyName;
+    }
+
+    public function getType(): Type
+    {
+        if ($this->propertyName == 'original') {
+          // See Drupal\Core\Entity\EntityStorageBase::doPreSave
+            return new ObjectType('Drupal\Core\Field\FieldItemListInterface');
```
mglaman: Isn't `original` a copy of the entity, so it should be an entity interface?

```
    // Load the original entity, if any.
    if ($id_exists && !isset($entity->original)) {
      $entity->original = $this->loadUnchanged($id);
    }
```

```
  /**
   * Loads an unchanged entity from the database.
   *
   * @param mixed $id
   *   The ID of the entity to load.
   *
   * @return \Drupal\Core\Entity\EntityInterface|null
   *   The unchanged entity, or NULL if the entity cannot be loaded.
   *
   * @todo Remove this method once we have a reliable way to retrieve the
   *   unchanged entity from the entity object.
   */
  public function loadUnchanged($id);
```

This should be `\Drupal\Core\Entity\EntityInterface`

src/Reflection/EntityFieldReflection.php
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return new ObjectType('Drupal\Core\Field\FieldItemListInterface');
```
mglaman: This is tricky.  This applies only to ContentEntityBase

```
  /**
   * Implements the magic method for getting object properties.
   *
   * @todo: A lot of code still uses non-fields (e.g. $entity->content in view
   *   builders) by reference. Clean that up.
   */
  public function &__get($name) {
    // If this is an entity field, handle it accordingly. We first check whether
    // a field object has been already created. If not, we create one.
    if (isset($this->fields[$name][$this->activeLangcode])) {
      return $this->fields[$name][$this->activeLangcode];
    }
    // Inline getFieldDefinition() to speed things up.
    if (!isset($this->fieldDefinitions)) {
      $this->getFieldDefinitions();
    }
    if (isset($this->fieldDefinitions[$name])) {
      $return = $this->getTranslatedField($name, $this->activeLangcode);
      return $return;
    }
    // Else directly read/write plain values. That way, non-field entity
    // properties can always be accessed directly.
    if (!isset($this->values[$name])) {
      $this->values[$name] = NULL;
    }
    return $this->values[$name];
  }
```

It could be a FieldItemListInterface or regular property. I guess we can handle unique properties one by one like `original`.

src/Reflection/FieldItemListPropertyReflection.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function getType(): Type
+    {
+        if ($this->propertyName == 'entity') {
+          // It was a EntityReferenceFieldItemList
+            return new ObjectType('Drupal\Core\Entity\FieldableEntityInterface');
+        }
+        if ($this->propertyName == 'target_id') {
+          // It was a EntityReferenceFieldItemList
+            return new IntegerType();
```
mglaman: This isn't true: it could be a machine name. And it's always a string regardless (string'd integer)

src/Reflection/FieldItemListPropertyReflection.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $names = ['entity', 'value', 'target_id'];
+        return in_array($propertyName, $names, true);
+    }
+
+    public function getType(): Type
+    {
+        if ($this->propertyName == 'entity') {
+          // It was a EntityReferenceFieldItemList
+            return new ObjectType('Drupal\Core\Entity\FieldableEntityInterface');
```
mglaman: We should be less specific and go EntityInterface as it could be a config entity reference

src/Reflection/FieldItemListPropertyReflection.php
```diff
@@ -0,0 +1,80 @@
...
+    public function __construct(ClassReflection $declaringClass, string $propertyName)
+    {
+        $this->declaringClass = $declaringClass;
+        $this->propertyName = $propertyName;
+    }
+
+    public static function canHandleProperty(string $propertyName): bool
+    {
+        $names = ['entity', 'value', 'target_id'];
+        return in_array($propertyName, $names, true);
```
mglaman: Should add a comment on "how to handle fields when this isn't true" like Address field.

src/Reflection/FieldItemListPropertyReflection.php
```diff
@@ -0,0 +1,80 @@
...
+        if ($this->propertyName == 'entity') {
+          // It was a EntityReferenceFieldItemList
+            return new ObjectType('Drupal\Core\Entity\FieldableEntityInterface');
+        }
+        if ($this->propertyName == 'target_id') {
+          // It was a EntityReferenceFieldItemList
+            return new IntegerType();
+        }
+        if ($this->propertyName == 'value') {
+            return new MixedType();
```
mglaman: Add a todo to provide some kind of resolving for this to provide a proper return type based on the assume field class.

hannsen:
Lol sorry, I wrote a review like a month ago, but forgot to submit it. Not used to Github Reviews

src/Reflection/EntityFieldReflection.php
hannsen: seems like a typo, because it returns FieldItemListInterface outside the if anyhow

src/Reflection/FieldItemListPropertyReflection.php
hannsen: Afaik it's always a string when you load the entity from the Database. Even booleans are just '0' and '1'

mglaman: Yeah, it's always a string unless you do `getValue` on the property itself and TypedData casts it.

src/Reflection/FieldItemListPropertyReflection.php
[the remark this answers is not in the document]
mglaman: This is true. But I think we can handle more verbose field handling in a follow up

src/Reflection/EntityFieldsViaMagicReflectionExtension.php
```diff
@@ -0,0 +1,43 @@
...
+    public function hasProperty(ClassReflection $classReflection, string $propertyName): bool
+    {
+        if ($classReflection->hasNativeProperty($propertyName)) {
+          // Let other parts of PHPStan handle this.
+            return false;
+        }
+
+        $reflection = $classReflection->getNativeReflection();
+        if ($reflection->implementsInterface('Drupal\Core\Entity\EntityInterface')) {
+            return true;
```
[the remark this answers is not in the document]
mglaman: Yeah. This is a doozy to try and handle. Documenting the fact it needs to be improved, because this PR adds some good benefit

mglaman:
🎉

PR #35, opened 2019-02-14
Replace getcwd() with global autoloaderInWorkingDirectory

### LINKED ISSUE #34: Breaks when using PHPStan globally
This is a regressions from #25. We need to determine the Drupal root based on the global/magical "autoloaderInWorkingDirectory" variable

```
        $finder = new DrupalFinder();
        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
```

mglaman:
Fixes #34

### DIFF
--- a/src/DependencyInjection/DrupalExtension.php
+++ b/src/DependencyInjection/DrupalExtension.php
@@ -52,7 +52,7 @@ class DrupalExtension extends CompilerExtension
     public function loadConfiguration(): void
     {
         $finder = new DrupalFinder();
-        $finder->locateRoot(getcwd());
+        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
         $this->drupalRoot = $finder->getDrupalRoot();
 
         $builder = $this->getContainerBuilder();
--- a/src/Drupal/Bootstrap.php
+++ b/src/Drupal/Bootstrap.php
@@ -62,7 +62,7 @@ class Bootstrap
     public function register(): void
     {
         $finder = new DrupalFinder();
-        $finder->locateRoot(getcwd());
+        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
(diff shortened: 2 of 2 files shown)


src/DependencyInjection/DrupalExtension.php
```diff
@@ -52,7 +52,7 @@ class DrupalExtension extends CompilerExtension
     public function loadConfiguration(): void
     {
         $finder = new DrupalFinder();
-        $finder->locateRoot(getcwd());
+        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
```
jibran: I don't think this is a correct fix. For global installation, we should expose the `--root` option just like drush.

No one installs Drupal globally so using some autoloader variable to guess Drupal root doesn't make sense.

mglaman: This is a PHPStan-ism. That global is what PHPStan uses to determine the target project's composer.json vs its own

mglaman: See https://github.com/phpstan/phpstan/blob/master/bin/phpstan

```
$autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
if (is_file($autoloaderInWorkingDirectory)) {
	require_once $autoloaderInWorkingDirectory;
}
$composerAutoloadFile = __DIR__ . '/../vendor/autoload.php';
if (!is_file($composerAutoloadFile)) {
	$composerAutoloadFile = __DIR__ . '/../../../autoload.php';
}
```

jibran: Requiring autoloader is different than finding Drupal root. As I said composer packages can be installed globally but installing Drupal globally doesn't make sense unless we fix https://www.drupal.org/project/drupal/issues/2910136 ;-)

mglaman: I'm confused, no one would need Drupal locally. This is if running phpstan-drupal from a global area. IE: a phar like https://github.com/mglaman/drupal-check

PR #41, opened 2019-03-23
Suggest installing deprecation rules as dev dependency

eiriksm:
Hola!

Testing this package, and just noticed what I find is a somewhat bad advice :)

One probably wants to install the deprecation rules package as a dev dependency (as one does with all other phpstan libraries)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ parameters:
 Add the deprecation rules to your Drupal project's dependencies
 
 ```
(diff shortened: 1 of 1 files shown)


mglaman:
Yep! Good catch, thanks.

eiriksm:
:+1: 

What a totally awesome package btw! Sorry I did not say that first in my PR body (I usually make a point of it). :D

Issue #46, opened 2019-03-26
Run phpstan under webroot


JTubex:
Currently, phpstan only works when running the command in the directory where Drupal is located.

Is it possible to run the command under the webroot? e.g.:
`www/vendor/bin/phpstan analyse www/modules/custom`

mglaman:
PHPStan needs to be able to load the autoloader for Drupal. So you must run PHPStan from the root of your Drupal project so it can detect the autoloader.

Now, I wonder if we could actually improve this:

```
        $finder = new DrupalFinder();
        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
        $this->drupalRoot = $finder->getDrupalRoot();
```

Instead of basing it off of `autoloaderInWorkingDirectory`, I wonder if we could determine the Drupal root based off of the provided path.  We would then need to include the autoloader if it was not. 

This is some of the magic in PHPStan

```
$autoloaderInWorkingDirectory = getcwd() . '/vendor/autoload.php';
if (is_file($autoloaderInWorkingDirectory)) {
	require_once $autoloaderInWorkingDirectory;
}

$composerAutoloadFile = __DIR__ . '/../vendor/autoload.php';
if (!is_file($composerAutoloadFile)) {
	$composerAutoloadFile = __DIR__ . '/../../../autoload.php';
}

require_once $composerAutoloadFile;
```

I'll try and experiment. But it might be possible to fix this by taking in the target path's folder and passing that to DrupalFinder.

mglaman:
Drupal Check https://github.com/mglaman/drupal-check does this. But that doesn't allow for customized PHPStan config, it started as mostly a helper for deprecation discovery

JTubex:
Reason i’m asking: We use GrumPHP (https://github.com/phpro/grumphp) to run some checks when committing. The commands are run from the directory where the .git directory is located. For us, this is under the webroot.

JTubex:
I created PR #50.

Instead of using  $GLOBALS['autoloaderInWorkingDirectory'], I used $GLOBALS['composerAutoloadFile'] which did the trick for me.

mglaman:
I'll begin reviewing the PR. I hate GitHub. It fractures discussions and code reviews.

mglaman:
Per my comments on the PR, I think the only fix is allowing an explicit parameter definition which specifies where Drupal's root is located and we detect the autoloader based off of that path.

JTubex:
[...]

PR #50, opened 2019-04-04
Make it possible to run phpstan under webroot

JTubex:
#46

### DIFF
--- a/src/DependencyInjection/DrupalExtension.php
+++ b/src/DependencyInjection/DrupalExtension.php
@@ -52,7 +52,7 @@ class DrupalExtension extends CompilerExtension
     public function loadConfiguration(): void
     {
         $finder = new DrupalFinder();
-        $finder->locateRoot(dirname($GLOBALS['autoloaderInWorkingDirectory'], 2));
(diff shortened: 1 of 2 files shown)


mglaman:
This is causing a failure:

```
0.23s$ ./vendor/bin/phpstan analyze web/core/modules/migrate_drupal --no-progress | grep -q "tests/fixtures" && false || true
Note: Using configuration file /home/travis/build/mglaman/drupal/phpstan.neon.
\Drupal::$container is not initialized yet. \Drupal::setContainer() must be called with a real container.
The command "./vendor/bin/phpstan analyze web/core/modules/migrate_drupal --no-progress | grep -q "tests/fixtures" && false || true" exited with 0.
```

mglaman:
The reason it uses the working directory definition is to support global or symlinked installs. You cannot always be sure the `composerAutoloadFile` is Drupal's autoloader.

Honestly I think the only real workaround is allowing explicit definition of the Drupal project's root as a parameter.

mglaman:
Sorry, HEAD is broken with that error!

PR #52, opened 2019-04-08
Update Bootstrap.php

mglaman:

Issue #51, opened 2019-04-06
Analyzing *.install: \Drupal::$container is not initialized yet


mglaman:
\Drupal::$container is not initialized yet. \Drupal::setContainer() must be called with a real container.

mglaman:
See #52.  Caused by `entity_test_update/entity_test_update.install` in `DbUpdatesTrait`

jibran:
I don't think this is totally fixed. If you use `\Drupal` anywhere in your install file you'll see this error. For example 

```- ./vendor/bin/phpstan analyze web/modules/contrib/token --no-progress || if (($? == 255)); then false; else true; 
```
will fail. I'll create a PR to show the fail.

jibran:
I created #66 but it didn't fail. I'll investigate more.

PR #66, opened 2019-05-02
Analyzing *.install still shows: \Drupal:: is not initialized yet

jibran:
Signed-off-by: Jibran Ijaz <jibran.ijaz@gmail.com>

This is a follow up of #51 and #52 which got reverted in #53. 

If `\Drupal` is used anywhere in `*.install` file it shows: 
```
\Drupal::$container is not initialized yet. \Drupal::setContainer() must be called with a real container.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ script:
 
   # Check "Cannot redeclare token_theme() due to blazy_test.module"
   - composer require drupal/token drupal/blazy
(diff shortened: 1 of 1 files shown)


mglaman:
I realized the testing was pretty broken. Hopefully this can be caught in PHPUNit as well

mglaman:
This should only occur _if_ `\Drupal::service` is invoked anywhere in a file when it is required. Such as `require $file` ends up executing code

jibran:
I think that is correct I'll have a look later today.

jibran:
Did you fix this in #68?

PR #68, opened 2019-05-07
Add graceful handling of container not being initialized

### LINKED ISSUE #67: Gracefully handle and report ContainerNotInitializedException
Reference: https://github.com/mglaman/drupal-check/issues/42

Currently, we skip known Drupal core modules which call `\Drupal::` outside of the scope of any function in their `.install` file.

```
                $ignored_install_files = ['entity_test', 'entity_test_update', 'update_test_schema'];
                if (!in_array($module_name, $ignored_install_files, true)) {
                    require $module_dir . '/' . $module_name . '.install';
                }
```

Turns out that the Examples module does this, too, in `tabledrag_example`.

```
module_load_include('inc', 'tabledrag_example', 'tabledrag_example.data');
```

The loading of modules should be wrapped in a try/catch for this exception and throw a custom scoped exception. This can help the end user understand _why_ this happened and how to fix it.

mglaman:
Fixes #67

### DIFF
--- a/src/Drupal/Bootstrap.php
+++ b/src/Drupal/Bootstrap.php
@@ -2,6 +2,7 @@
 
 namespace PHPStan\Drupal;
 
+use Drupal\Core\DependencyInjection\ContainerNotInitializedException;
 use DrupalFinder\DrupalFinder;
 use Nette\Utils\Finder;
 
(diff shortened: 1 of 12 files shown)


src/Drupal/Bootstrap.php
```diff
@@ -223,4 +224,17 @@ protected function loadExtension(Extension $extension): void
...
         }
     }
+
+    protected function loadAndCatchErrors(string $path): void
+    {
+        try {
+            require $path;
+        } catch (ContainerNotInitializedException $e) {
+            // This can happen when drupal_get_path or drupal_get_filename are used outside of the scope of a function.
+            @trigger_error("$path invoked the Drupal container outside of the scope of a function or class method. It was skipped for analysis.", E_USER_WARNING);
```
mglaman: proof of concept catching these to be replayed in drupal-check or another runner.

PR #69, opened 2019-05-07
Detection of deprecated constants

### LINKED ISSUE #58: Deprecated global constants are not discovered
See https://github.com/drupalcommerce/commerce/commit/e07e89819c21b472b7938637af003991b8fe6fad

mglaman:
Fixes #58

### DIFF
--- a/src/Rules/Deprecations/AccessDeprecatedConstant.php
+++ b/src/Rules/Deprecations/AccessDeprecatedConstant.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types = 1);
+
+namespace PHPStan\Rules\Deprecations;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Broker\Broker;
+use PHPStan\Rules\RuleLevelHelper;
+
+class AccessDeprecatedConstant implements \PHPStan\Rules\Rule
+{
+    /** @var Broker */
+    private $broker;
+    /** @var RuleLevelHelper */
+    private $ruleLevelHelper;
+    public function __construct(Broker $broker, RuleLevelHelper $ruleLevelHelper)
+    {
+        $this->broker = $broker;
+        $this->ruleLevelHelper = $ruleLevelHelper;
+    }
+
+    public function getNodeType(): string
+    {
+        return Node\Expr\ConstFetch::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (DeprecatedScopeHelper::isScopeDeprecated($scope)) {
+            return [];
+        }
+
+        // comment is not resolved.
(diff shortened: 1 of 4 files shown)


tests/src/DeprecationRulesTest.php
```diff
@@ -19,19 +19,27 @@ public function testDeprecationRules(string $path, int $count, array $errorMessa
...
-                'Call to deprecated method url() of class Drupal.'
-            ]
-        ];
-        yield [
-            __DIR__ . '/../fixtures/drupal/core/lib/Drupal/Core/Entity/EntityManager.php',
-            1,
-            [
-                'Class Drupal\Core\Entity\EntityManager implements deprecated interface Drupal\Core\Entity\EntityManagerInterface.'
+                'Call to deprecated constant DATETIME_STORAGE_TIMEZONE.',
+                'Call to deprecated constant DATETIME_DATE_STORAGE_FORMAT.',
```
goba: These should be failing now as we should now get the deprecation message as well.

src/Rules/Deprecations/AccessDeprecatedConstant.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (DeprecatedScopeHelper::isScopeDeprecated($scope)) {
+            return [];
+        }
+
+        // comment is not resolved.
+        $comment = $node->getDocComment();
```
goba: Why do we even need the doc comment here? Is it to decide from that if the constant was deprecated?

src/Rules/Deprecations/AccessDeprecatedConstant.php
```diff
@@ -0,0 +1,39 @@
...
+        $constnatName = $this->broker->resolveConstantName($node->name, $scope);
```
goba: Typo constnatName :)

src/Rules/Deprecations/AccessDeprecatedConstant.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function getNodeType(): string
+    {
+        return Node\Expr\ConstFetch::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (DeprecatedScopeHelper::isScopeDeprecated($scope)) {
+            return [];
```
goba: Is this to avoid reporting deprecated things in already deprecated scopes? Yay!

mglaman: yep!

mglaman:
Problems. nikic/php-parser doesn't seem to pick up comments on global constants.

Screen Shot 2019-07-29 at 5 17 04 PM

PR #70, opened 2019-05-09
Fix root drush directory path detection

hgajjar:
Drush class is at `drush/drush/lib/Drush/Drush.php`, so `dirname()` with level `2` would give `drush/drush/lib` and then it tried to append `/includes` to it, but include directory is at `drush/drush/includes`. So, setting it to level `3` fixes this issue.


hgajjar:
oh, seems like this issue is with old version of drush only, closing 😄

mglaman:
Do we need to try and support Drush 8 and 9? Because we can

hgajjar:
@mglaman that would be great!

hgajjar:
@mglaman PR is ready #72

PR #75, opened 2019-06-11
Fixes Drupal core can't be find by DrupalTestBot

attrib:
I'm not sure why the dirname was cut by 2 levels, but this make trouble with DrupalTestBot.

See https://www.drupal.org/project/upgrade_status/issues/3050653

It seems what DrupalTestBot does is something along the line
```
git clone drupal
composer install
install upgrade_status
```

In this case the autoload.php is in the root dir:
```
core
vendor
autoload.php
```
all in the same directory.

while when I use composer create project the drupal root is in a sub directory and the vendor one level above and then the current code works.

### DIFF
--- a/src/DependencyInjection/DrupalExtension.php
+++ b/src/DependencyInjection/DrupalExtension.php
@@ -66,7 +66,7 @@ class DrupalExtension extends CompilerExtension
         if ($config['drupal_root'] !== '' && realpath($config['drupal_root']) !== false && is_dir($config['drupal_root'])) {
             $start_path = $config['drupal_root'];
         } else {
-            $start_path = dirname($GLOBALS['autoloaderInWorkingDirectory'], 2);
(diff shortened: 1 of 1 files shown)


mglaman:
So this would break Composer installs. There's an issue to improve all of this. I think I opened a PR. On mobile and can't see quickly

mglaman:
Lining to #71

goba:
@mglaman When I raised it breaking composer installs, @attrib explained the code goes into the directory to try and find the right subfolder anyway, and he tried on both his composer based on install and the git checkout install, and it worked for him.

mglaman:
Sorry, it doesn't break anything. If the project root is an additional directory above, the DrupalFinder will discover it anyways.

mglaman:
🤔 I wish I could re-trigger CircleCI for this, as it has better test coverage.

mglaman:
🙌 found the right payload to trigger.

mglaman:
@attrib thanks!

PR #77, opened 2019-06-13
Prevent files from being required twice

attrib:
Using phpstan 0.11.3 and dev of phpstan-drupal (after #75 ) and the drupal upgrade_status module dev version you receive an error "Drupal already defined"

Issue for this is that `require` and not `require_once` is used.

I already mentioned this in https://github.com/mglaman/phpstan-drupal/pull/71#issuecomment-501213378

After talking with Gábor Hojtsy, we decided its maybe better to create a separate PR for this instead of only in this comment.

As you already mentioned in https://github.com/mglaman/phpstan-drupal/pull/71#issuecomment-501253478 its a good question if this `require` is needed at all, as composer should load it anyways.

Maybe get this in now and do the refactoring/removing in #71 ?
In this way we should get the drupal tests running and then can test the refactoring better?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,43 @@ jobs:
           command: |
             cd /tmp/drupal
             ./vendor/bin/phpstan analyze web/core/modules/dynamic_page_cache --debug
+  test_upgrade_status:
+    <<: *defaults
+    steps:
+      - start-project
+      - run:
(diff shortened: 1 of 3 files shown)


mglaman:
My concern is the fact we don't have test failures here. There's a lot of magic happening and I want to try and have test coverage

mglaman:
@attrib can you link to a test failure?

PR #85, opened 2019-08-14
Exclude build directory from discovery

mxr576:
Well, this a PR that kinda solves our specific problem and it is related to:
https://github.com/mglaman/drupal-check/pull/98 and https://www.drupal.org/project/drupal/issues/2943172

I understand this PR is probably not going to be merged as-is but I hope it starts a conversation about how this problem can be solved in a maintainable way (and can use this PR as a patch with Composer as a temporary fix our builds ;P)

I see that this library basically has a copy-paste of Drupal core's extension discovery with a major difference. It does not retrieve the ignored files list from Settings or from a configuration parameter: https://github.com/drupal/core/blob/8.7.6/lib/Drupal/Core/Extension/ExtensionDiscovery.php#L427

### DIFF
--- a/src/Drupal/ExtensionDiscovery.php
+++ b/src/Drupal/ExtensionDiscovery.php
@@ -334,7 +334,7 @@ class ExtensionDiscovery
         // this to not check for files in common special-purpose directories. For
         // example, node_modules and bower_components. Ignoring irrelevant
         // directories is a performance boost.
-        $ignore_directories = ['node_modules', 'bower_components'];
+        $ignore_directories = ['node_modules', 'bower_components', 'build'];
 
         // Filter the recursive scan to discover extensions only.
         // Important: Without a RecursiveFilterIterator, RecursiveDirectoryIterator


mglaman:
The setting should be moved to a extension.neon parameter.

mxr576:
My other idea was using environment variables for this so the list could be extended by the runtime environment (like Docker) easily.

src/Drupal/ExtensionDiscovery.php
```diff
@@ -334,7 +334,7 @@ protected function scanDirectory($dir): array
         // this to not check for files in common special-purpose directories. For
         // example, node_modules and bower_components. Ignoring irrelevant
         // directories is a performance boost.
-        $ignore_directories = ['node_modules', 'bower_components'];
+        $ignore_directories = ['node_modules', 'bower_components', 'build'];
```
mglaman: This should be a service parameter in the extension.neon file

mglaman:
@mxr576 is this still a problem? The fix would be adding a parameter in `extension.neon` to make this configurable.

mxr576:
Well, we switch the way we test modules in isolation, so currently it is not a blocker, but probably still a good idea making this configurable, extendible.

mglaman:
I'm going to close this. It hasn't been reported by many others. I don't want to add a parameter that could be abused as a way to avoid scanning certain modules in unsupported use cases.

PR #86, opened 2019-10-14
Make sure the service providers can be found with cold caches

### LINKED ISSUE #31: Extension service provider classes are not autoloaded.
The pattern is camel case of the extension with ServiceProvider appended.

```
 ------ ------------------------------------------------------------------------- 
  Line   src/EntityReferenceRevisionsServiceProvider.php                          
 ------ ------------------------------------------------------------------------- 
         Class                                                                    
         Drupal\entity_reference_revisions\EntityReferenceRevisionsServiceProvid  
         er was not found while trying to analyse it - autoloading is probably    
         not configured properly.                                                 
 ------ ------------------------------------------------------------------------- 
```

eiriksm:
This fixes #31 

So, I do not know how the piece of code is supposed to actually work, since it used to just set service providers in an array and never use them for anything. 

The point is, it would even work had the autoloader already mapped the service providers, because then they would be registered by class_exists. But since this runs before the registering of classes in Bootstrap::addModuleNamespaces service providers are never found.

I have now solved this by assigning all potential module service provider to the drupalServiceMap parameter. I am pretty sure this could be factored out to its own factory, and actually check when things are needed, but this solves the problem and should not give any side effects.

The most annoying part of the bug is that class definitions are cached, so after you run the command once, classes are found. Which is not something we want to do in a CI environment, where we use phpstan.

Anyway, here are steps to reproduce:

- Create a module with a service provider. In my case this is called `JsonapiSafeDefaultsServiceProvider.php`
- Make sure the phpstan cache is cleared. In my case right now this means deleting this folder: `rm -rf /var/folders/50/8lm05cmd1h548wdj0953sr000000gn/T/phpstan/`. Your folder might differ, and can be found by running this command: `php -r "echo sys_get_temp_dir();"`
- Have this package and others installed, and a pretty common setup with phpstan.neon
- Run the command on the folder where you created a service provider.

Sample output:

```
eiriks-iMac:my-project eirikmorland$ rm -rf /var/folders/50/8lm05cmd1h548wdj0953sr000000gn/T/phpstan/
eiriks-iMac:my-project eirikmorland$ ./vendor/bin/phpstan analyse drupal/modules/nymedia/jsonapi_safe_defaults/ --level=max
Note: Using configuration file /Users/xxx/phpstan.neon.
 4/4 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   src/JsonapiSafeDefaultsServiceProvider.php                                                                                                                         
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
         Class Drupal\jsonapi_safe_defaults\JsonapiSafeDefaultsServiceProvider was not found while trying to analyse it - autoloading is probably not configured properly.  
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

 ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 6 errors                                                                                                 
                                                                                                                        

eiriks-iMac:my-project eirikmorland$ ./vendor/bin/phpstan analyse drupal/modules/nymedia/jsonapi_safe_defaults/ --level=max
Note: Using configuration file /Users/xxx/phpstan.neon.
 4/4 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ -------------------------------------------- 
  Line   src/JsonapiSafeDefaultsServiceProvider.php  
 ------ -------------------------------------------- 
  18     If condition is always true.                
 ------ -------------------------------------------- 

As you can see, it works as expected after warming up the cache. Which is not very desirable :)

With this PR applied, everything works like expected with and without cache

### DIFF
--- a/src/DependencyInjection/DrupalExtension.php
+++ b/src/DependencyInjection/DrupalExtension.php
@@ -125,9 +125,11 @@ class DrupalExtension extends CompilerExtension
             $name = "{$camelized}ServiceProvider";
             $class = "Drupal\\{$moduleName}\\{$name}";
 
-            if (class_exists($class)) {
-                $serviceClassProviders[$moduleName] = $class;
-            }
+            $serviceClassProviders[$moduleName] = $class;
+            $serviceId = "service_provider.$moduleName.service_provider";
+            $builder->parameters['drupalServiceMap'][$serviceId] = [
+                'class' => $class,
+            ];
         }
 
         foreach ($serviceYamls as $extension => $serviceYaml) {

### CODE REVIEW

File: src/DependencyInjection/DrupalExtension.php [hunk 1]
```diff
@@ -125,9 +125,11 @@ public function loadConfiguration(): void
             $name = "{$camelized}ServiceProvider";
             $class = "Drupal\\{$moduleName}\\{$name}";
 
-            if (class_exists($class)) {
-                $serviceClassProviders[$moduleName] = $class;
-            }
+            $serviceClassProviders[$moduleName] = $class;
+            $serviceId = "service_provider.$moduleName.service_provider";
+            $builder->parameters['drupalServiceMap'][$serviceId] = [
```
mglaman: so the variable`$serviceClassProviders` is never actually used. We could use that to bulk stuff in the `drupalServiceMap`.

mglaman:
I wonder if improving the extension's integration with PHPStan would just help, in general https://github.com/mglaman/phpstan-drupal/pull/71

src/DependencyInjection/DrupalExtension.php
eiriksm: It is not, as mentioned in the beginning. Just left it here to avoid doing several things at once

eiriksm:
>So \PHPStan\Drupal\ServiceMapFactory::create receives the array of services and pushes it into \PHPStan\Drupal\ServiceMap.
>
>I don't know how this allows it to work.

It allows it to work by having the services in the bootstrap part, where we register namespaces in the autoloader.

eiriksm:
Is there anything more i can do to get this commited? :)

mglaman:
Forgot to merge 🙃  was trying to write a test elsewhere and forgot.

PR #95, opened 2019-12-06
Mark conflicts

mglaman:


composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "phpstan/phpstan": "^0.11",
```
goba: Per https://getcomposer.org/doc/articles/versions.md#tilde-version-range-, 

> The ~ operator is best explained by example: ~1.2 is equivalent to >=1.2 <2.0.0, while ~1.2.3 is equivalent to >=1.2.3 <1.3.0

So specifying tilde 0.11 will still let it update to 0.12, 0.13, etc. until it reaches 1.0 which it will NOT update to. However if you specify tilde 0.11.0, then it will not update to 0.12.

Adding a conflict on 0.12 would only work as long as there is not a 0.13, which tilde 0.11 still allows to update to.

ryanaslett:
The ^ vs ~ might be moot because I believe that composer treats anything that is *below* 1.0.0 as a pure dev release, and the constraints may not matter in that case.

goba:
> The ^ vs ~ might be moot because I believe that composer treats anything that is _below_ 1.0.0 as a pure dev release, and the constraints may not matter in that case.

I did not find that documented. On the other hand 

https://semver.mwl.be/#!?package=phpstan%2Fphpstan&version=~0.11.0&minimum-stability=stable shows 0.12 will not be an option that satisfies that while for https://semver.mwl.be/#!?package=phpstan%2Fphpstan&version=~0.11&minimum-stability=stable 0.12 will be satisfactory. If that is true for composer itself not just the tester, then we should go with `~0.11.0` IMHO.

mglaman:
@goba thanks, I made modifications to use `~0.11.0` require constraints and moved conflict to `0.12`. ideally this will be resolved before there's a 0.13

mglaman:
I don't know, I can't get it to test properly. But I'm going to merge and hope that does it

PR #97, opened 2019-12-10
Add all the deprected core constants

goba:
Includes the 5 deprecated class/interface constants commented out. I don't think we need those here as they should be found already(?) Otherwise this is complete list according to api.drupal.org.


src/Rules/Deprecations/AccessDeprecatedConstant.php
```diff
@@ -37,12 +37,71 @@ public function processNode(Node $node, Scope $scope): array
...
+            'USER_REGISTER_VISITORS_ADMINISTRATIVE_APPROVAL' => 'Deprecated in drupal:8.3.0 and is removed from drupal:9.0.0. Use \Drupal\user\UserInterface::REGISTER_VISITORS_ADMINISTRATIVE_APPROVAL instead.',
+            // 'BlockInterface::BLOCK_LABEL_VISIBLE' => 'Deprecated in drupal:8.3.0 and is removed from drupal:9.0.0. Use \Drupal\Core\Block\BlockPluginInterface::BLOCK_LABEL_VISIBLE.',
+            // 'BlockInterface::BLOCK_REGION_NONE' => 'Deprecated in drupal:8.?.? and is removed from drupal:9.0.0.',
+            // 'WorkspaceInterface::DEFAULT_WORKSPACE' => 'Deprecated in drupal:8.8.0 and is removed from drupal:9.0.0. Use \Drupal\workspaces\WorkspaceManager::hasActiveWorkspace() instead.',
+            // 'PluralTranslatableMarkup::DELIMITER' => 'Deprecated in drupal:8.7.0 and is removed from drupal:9.0.0. Use Drupal\Component\Gettext\PoItem::DELIMITER instead.',
+            // 'FieldMigration::PLUGIN_METHOD' => 'Deprecated in drupal:8.7.0 and is removed from drupal:9.0.0. Use the migrate_drupal.field_discovery service instead. See https://www.drupal.org/node/3006076.',
```
mglaman: You know better than I why this is commented out, so I trust it, I'm more so curious -- why?

goba: These were the only non-global constants on api.drupal.org. I had a hunch that these would be found as there should be no PHP tokenizer problems associating the phpdoc to class constants.... I did not check though before I submitted this, so just left them there.

But indeed phpstan already finds these, proof of BlockInterface::BLOCK_REGION_NONE at https://dev.acquia.com/drupal9/deprecation_status/errors/271 :)

So we could just remove these and not commit them :)

PR #99, opened 2019-12-14
Small PR for PHPStan 0.12 upgrade

ondrejmirtes:


extension.neon
```diff
@@ -12,13 +12,14 @@ parameters:
...
 	drupal:
+		drupal_root: null
 		entityTypeStorageMapping:
 			node: Drupal\node\NodeStorage
 			taxonomy_term: Drupal\taxonomy\TermStorage
 			user: Drupal\user\UserStorage
 parametersSchema:
 	drupal: structure([
-		drupal_root: string()
+		drupal_root: schema(string(), nullable())
```
mglaman: :) thank you

src/Drupal/DrupalAutoloader.php
```diff
@@ -63,8 +63,8 @@ class DrupalAutoloader
     public function register(Container $container): void
     {
         $startPath = null;
-        $drupalParams = $container->hasParameter('drupal') ? $container->getParameter('drupal') : [
-          'drupal_root' => null,
+        $drupalParams = [
+          'drupal_root' => $container->getParameter('drupal'),
```
mglaman: drupal_root is a param under drupal.  but with the schema, I shouldn't have to worry about missing data, correct?

mglaman:
Thank you @ondrejmirtes !

src/Drupal/DrupalAutoloader.php
ondrejmirtes: Yes, it will always have this parameter.

PR #105, opened 2020-01-13
EntityReferenceItem target_id can be integer

el7cosmos:


src/Reflection/FieldItemListPropertyReflection.php
```diff
@@ -57,7 +57,7 @@ public function getWritableType(): Type
             return new ObjectType('Drupal\Core\Entity\EntityInterface');
         }
         if ($this->propertyName === 'target_id') {
-            return new StringType();
+            return new UnionType([new StringType(), new IntegerType()]);
```
mglaman: Technically they could. But when loaded from the database they are always a string. Even when fetching an entity ID via `->id()` it will return a string unless you have implemented casting.

el7cosmos: The problem is when I do casting with return type PHPstan will fail, for example
```php
function (ContentEntityInterface $entity): int {
  return $entity->get('test')->target_id;
}
```
above won't work
```php
function (ContentEntityInterface $entity): int {
  return (int) $entity->get('test')->target_id;
}
```
while this works, but ended up with double casting, one in function return type, and other in return statement.

What do you think?

mglaman: That's what I was checking as I started doing the same thing 😄 I wanted to just make sure.

mglaman: I'd like to try and work on a test to make sure this is covered. Do you have the setting checked where maintainers can push to the PR branch?

el7cosmos: Yes I have

mglaman: Wait, with `declare(strict_types = 1);` at the top of the file, the following passes?

```
function (ContentEntityInterface $entity): int {
  return $entity->get('test')->target_id;
}
```

That should fail since it's a string. 🤔 in all of my code I _had_ to double cast.

el7cosmos: fair enough, i think this is general problem, following also fail
```php
function test(): int {
  return '1';
;
```

mglaman:
@el7cosmos wait, this is still valid.

el7cosmos:
@mglaman what your suggestion?

src/Reflection/FieldItemListPropertyReflection.php
mglaman: I think it's still worth adding this, do you agree? Or is it better to keep it explicit until Drupal core changes how it returns values?

mglaman:
@el7cosmos I was debating if it is worth adding. But if Drupal core returns strings from the database regardless, maybe we should assume that. Drupal's data type API doesn't even cast values unless you invoke `castedValue`. So maybe you're right and we should keep it as is, which will help people understand the values Drupal returns.

el7cosmos:
@mglaman I personally think we keep it as is for now

PR #109, opened 2020-01-21
Fixes extension autoloader to behave like Drupal core

mradcliffe:
See [drupal-check #124](https://github.com/mglaman/drupal-check/issues/124).

The autoloader infers that the behavior should match [drupal_phpunit_get_extension_namespaces](https://api.drupal.org/api/drupal/core%21tests%21bootstrap.php/function/drupal_phpunit_get_extension_namespaces/8.9.x), but the logic behind this does not match that function.

This removes the hard-coded namespaces and only adds the two necessary namespaces for extensions:

1. `\Drupal\module_name` -> module_root/src
2. `\Drupal\Tests\module_name` -> module_root/tests/src

I wasn't sure about tests so I added a module that implement a couple of test namespaces, but I didn't add all the current possibilities.

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -234,24 +234,12 @@ class DrupalAutoloader
             $module_dir = $this->drupalRoot . '/' . $module->getPath();
             $this->namespaces["Drupal\\$module_name"] = $module_dir . '/src';
 
-            // @see drupal_phpunit_get_extension_namespaces
+            // Extensions can have a \Drupal\Tests\extension namespace for test cases, traits, and other classes such
+            // as those that extend \Drupal\TestSite\TestSetupInterface.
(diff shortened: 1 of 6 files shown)


src/Drupal/DrupalAutoloader.php
```diff
@@ -234,24 +234,12 @@ protected function addModuleNamespaces(): void
...
-                    }
-
-                    // Extensions can have a \Drupal\extension\Traits namespace for
-                    // cross-suite trait code.
-                    $trait_dir = $module_test_dir . '/Traits';
-                    if (is_dir($trait_dir)) {
-                        $this->namespaces["Drupal\\Tests\\$module_name\\Traits"] = $trait_dir;
-                    }
-                }
+                $this->namespaces["Drupal\\Tests\\$module_name"] = $module_test_dir;
```
mglaman: 😄 much simpler

PR #110, opened 2020-01-21
Fix type inspection for app.root service

tstoeckler:
First of all, thanks for this great project, I started toying around with it recently and am liking it a lot. :+1: 

I found a quibble, though, which is what this PR is for:
I could not find a way with the current inspection to do something like:
```php
$app_root = $this->container->get('app.root');
$some_path = $app_root . '/core/includes/install.inc';
```

That would result in:

> Binary operation "." between SplString and '/core/incl…' results in an error.

Casting `$app_root` to a string is also not allowed, so I was pretty much out of options as far as I could tell.

This seemed to be a sensible way to fix it, but I'm certainly open to other suggestions.

### DIFF
--- a/src/Type/ServiceDynamicReturnTypeExtension.php
+++ b/src/Type/ServiceDynamicReturnTypeExtension.php
@@ -11,6 +11,7 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Constant\ConstantBooleanType;
+use Psr\Container\ContainerInterface;
 
 class ServiceDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
@@ -26,7 +27,7 @@ class ServiceDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtens
 
     public function getClass(): string
     {
-        return 'Symfony\Component\DependencyInjection\ContainerInterface';
+        return ContainerInterface::class;
     }
 
     public function isMethodSupported(MethodReflection $methodReflection): bool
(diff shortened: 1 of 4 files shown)


src/Type/ServiceDynamicReturnTypeExtension.php
```diff
@@ -55,6 +55,12 @@ public function getTypeFromMethodCall(
         if ($methodReflection->getName() === 'get') {
             $service = $this->serviceMap->getService($serviceId);
             if ($service instanceof DrupalServiceDefinition) {
+                // Work around Drupal misusing the SplString class for string
+                // pseudo-services such as 'app.root'.
+                // @see https://www.drupal.org/project/drupal/issues/3074585
+                if ($service->getClass() === 'SplString') {
+                    return new StringType();
+                }
```
mglaman: 👍 makes sense!

tstoeckler:
Thanks that would be great! Yes, the checkbox should be checked.

mglaman:
@tstoeckler actually, it seems like part of the problem is in PHPStan that it doesn't believe `SplString` can cast to string, which works just fine. Correct?

mglaman:
For instance,

```yaml
  info_parser:
    class: Drupal\Core\Extension\InfoParser
    arguments: ['@app.root']
```

\Drupal\Core\Extension\InfoParserDynamic

```
  public function __construct(string $app_root = NULL) {
```

So it accepts it as a string. But.. I just realized that file doesn't have strict types declared, either.

mglaman:
Ah, because it calls a factory which returns a string.. and you need to define a class at least?

```
  /**
   * Gets the app root.
   *
   * @return string
   */
  public function get() {
    return $this->drupalKernel->getAppRoot();
  }
```

mglaman:
```
Psy Shell v0.9.12 (PHP 7.3.12 — cli) by Justin Hileman
>>> print gettype(\Drupal::service('app.root'));
string⏎
=> 1
```

Okay, it does return a string type and not SplString.

mglaman:
It actually _is_ correct to complain. If you try to do `$test = (new SplString('test')) . 'test';` it'll error out because SplString does not have `__toString`. 

I added a test, to verify coverage. Thanks @tstoeckler ! 🙌

tstoeckler:
Awesome, thanks for the quick turnaround. :tada: 

Ahh, I didn't know that about `SplString` thanks for explaining. I just assumed that that would work.

PR #115, opened 2020-02-13
See https://www.drupal.org/node/3113284.

mpdonadio:
Per @Berdir in https://www.drupal.org/project/drupal/issues/2902895#comment-13461286,

> @catch and @alexpott agreed in Slack to postpone removing the constant to D10 as it seems like a lot of work is left that close to the first beta deadline. We can keep working on its removal in core but should also do an issue to change the deprecation message. And I guess remove it again from the hardcoded list in phpstan-drupal?

See https://www.drupal.org/node/3113284 for the issue w/ the patch


mglaman:
Thanks! I'll work on tagging a release of phpstan-drupal.

PR #129, opened 2020-06-09
Replace deprecated autoload_files parameter

brambaud:
`autoload_files` parameter has been deprecated in [phpstan 0.12.26]( https://github.com/phpstan/phpstan/releases/tag/0.12.26).

```bash
$ ./bin/phpstan --version
PHPStan - PHP Static Analysis Tool 0.12.26
$ ./bin/phpstan analyse
Note: Using configuration file /app/phpstan.neon.dist.
⚠️  You're using a deprecated config option autoload_files. ⚠️️

You might not need it anymore - try removing it from your
configuration file and run PHPStan again.

If the analysis fails, there are now two distinct options
to choose from to replace autoload_files:
1) scanFiles - PHPStan will scan those for classes and functions
   definitions. PHPStan will not execute those files.
2) bootstrapFiles - PHPStan will execute these files to prepare
   the PHP runtime environment for the analysis.

Read more about this in PHPStan's documentation:
https://phpstan.org/user-guide/discovering-symbols
```

As far as I understood in order to keep the same behavior a quick way could be to replace the deprecated `autoload_files` by `bootstrapFiles`.


brambaud:
Seemed too good to be true to have the tests passed on the first try ^^

Will have to check more deeply

PR #135, opened 2020-07-26
Flag deprecated hook usages as errors

### LINKED ISSUE #126: Deprecated hooks not flagged
I'm using `phpstan-drupal` with `phpstan-deprecation-modules`, and it doesn't appear to be flagging up usage of deprecated hooks. Specifically, I have an implementation of `hook_search_api_results_alter` in a custom module that is being scanned, and the implementation doesn't get flagged even though the hook is marked as deprecated. This would be a super useful feature for code checking for D9.

mglaman:
Fixes #126﻿

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,8 @@
         "phpunit/phpunit": "^7.5",
         "phpstan/phpstan-deprecation-rules": "~0.12.0",
         "composer/installers": "^1.6",
-        "drupal/core-recommended": "^8.8@alpha",
-        "drush/drush": "^9.6"
+        "drupal/core-recommended": "^8.9",
+        "drush/drush": "^10"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
--- a/src/Rules/Deprecations/HookDeprecated.php
+++ b/src/Rules/Deprecations/HookDeprecated.php
@@ -61,6 +61,5 @@ class HookDeprecated implements Rule
             }
         }
         return $errors;
-
     }
 }
--- a/src/Rules/Deprecations/HookDeprecated.php
+++ b/src/Rules/Deprecations/HookDeprecated.php
@@ -7,6 +7,7 @@ use PhpParser\Node\Identifier;
 use PHPStan\Analyser\Scope;
 use PHPStan\Broker\Broker;
 use PHPStan\Rules\Rule;
+use PHPStan\ShouldNotHappenException;
 
 class HookDeprecated implements Rule
 {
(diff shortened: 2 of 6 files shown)


mglaman:
For the life of me I couldn't figure out how to get the values, the answer was in #128 all along

```
            $module_arg = $node->args[0]->value->value;
            $type_arg = $node->args[1]->value->value;
            $name_arg = !empty($node->args[2]) ? $node->args[2] : null;
```

src/Rules/Deprecations/HookDeprecated.php
```diff
@@ -0,0 +1,54 @@
...
+            throw new ShouldNotHappenException('alterDeprecated was called with missing arguments.');
+        }
+        $arg_description = $args[0]->value;
+        assert($arg_description instanceof Node\Scalar\String_);
+        $arg_type = $args[1]->value;
+        assert($arg_type instanceof Node\Scalar\String_);
+
+        $deprecation_description = $arg_description->value;
+        $hook_type = $arg_type->value;
+        return [sprintf(
```
mglaman, line 48: This should not return anytime `alterDeprecated` is called. It should check the `Broker` to see if any modules implement the hook.

mglaman: The one error should be `phpstan_fixtures_simpletest_alter`

mglaman: It should make an error for each implemented hook.

composer.json
```diff
@@ -22,8 +22,8 @@
         "phpunit/phpunit": "^7.5",
         "phpstan/phpstan-deprecation-rules": "~0.12.0",
         "composer/installers": "^1.6",
-        "drupal/core-recommended": "^8.8@alpha",
-        "drush/drush": "^9.6"
+        "drupal/core-recommended": "^8.9",
+        "drush/drush": "^10"
```
mglaman, line 26: - [ ] Revert this change, it caused Drush checks to fail, not needed until D9 support added.

src/Rules/Deprecations/HookDeprecated.php
```diff
@@ -0,0 +1,66 @@
...
+            return [];
+        }
+        $args = $node->args;
+        if (count($args) < 3) {
+            throw new ShouldNotHappenException('alterDeprecated was called with missing arguments.');
+        }
+        $arg_description = $args[0]->value;
+        assert($arg_description instanceof Node\Scalar\String_);
+        // @todo this could be an array of hook types.
+        $arg_type = $args[1]->value;
```
mglaman, line 42: - [ ] Support an array for argument
- [ ] Test a call which uses an array

tests/src/DeprecationRulesTest.php
```diff
@@ -47,5 +47,17 @@ public function dataDeprecatedSamples(): \Generator
                 'Call to deprecated constant DATETIME_DATE_STORAGE_FORMAT: Deprecated in drupal:8.5.0 and is removed from drupal:9.0.0. Use \Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface::DATE_STORAGE_FORMAT instead.',
             ]
         ];
+        yield [
+            __DIR__ . '/../fixtures/drupal/core/modules/simpletest/src/TestDiscovery.php',
```
mglaman, line 51: This highlights one problem: we can only detect usage of deprecated hooks if the code which invokes the hook is analyzed.

mglaman:
Closing, but keeping branch. The approach did not work. It only worked the hook is dispatched.

Issue #143, opened 2020-09-13
 Reflection error: Drupal\Tests\PhpunitCompatibilityTrait not found


MPParsley:
**How is drupal-check installed?**

drupal-check is installed as a dependency to my project

**Environment:**
 - OS: macOS
 - PHP Version: 7.3
 - Drupal core: 9.0.x

**Describe the bug**
Reflection error occurs when using multiple arguments to test different files.
$ vendor/bin/drupal-check --deprecations --no-progress --drupal-root=../drupal --exclude-dir=vendor cookiebot_consent.module src/Event/LibrariesEvent.php

This issue was partially fixed in https://github.com/phpstan/phpstan/issues/3522 but still occurs when using different files as arguments.

cookiebot_consent.module
```
  $event = new LibrariesEvent($extension);
```

src/Event/LibrariesEvent.php
```
use Symfony\Contracts\EventDispatcher\Event;

class LibrariesEvent extends Event {
```

Symfony\Contracts\EventDispatcher\Event
```
namespace Symfony\Contracts\EventDispatcher;

use Psr\EventDispatcher\StoppableEventInterface;

if (interface_exists(StoppableEventInterface::class)) {
    class Event implements StoppableEventInterface
    {
    }
} else {
    class Event
    {
    }
}
```

**Console output**
```
 ------ -------------------------------------------------------------------- 
  Line   cookiebot_consent.module                                            
 ------ -------------------------------------------------------------------- 
         Reflection error: Psr\EventDispatcher\StoppableEventInterface not   
         found.                                                              
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ------ -------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------- 
  Line   src/Event/LibrariesEvent.php                                        
 ------ -------------------------------------------------------------------- 
         Reflection error: Psr\EventDispatcher\StoppableEventInterface not   
         found.                                                              
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  52     Reflection error: Psr\EventDispatcher\StoppableEventInterface not   
         found.                                                              
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
 ------ -------------------------------------------------------------------- 
```

mglaman:
#144 fixes the tests here and uncovers this bug

mglaman:
So this path is covered in the autoloader provided by PHPStan Drupal.

```
[...]

PR #149, opened 2020-10-12
Remove PHPStan conflict

### LINKED ISSUE #152: phpstan-drupal should not conflict with phpstan/phpstan versions
Problem in composer.json:
```
    "conflict": {
        "phpstan/phpstan": "^0.12.43"
    },
```
Why is that necessary? It blocks me from upgrading PHPStan to any higher version.

mglaman:
Tries to see if latest PHPStan has anything for fixing #143 

Fixes #152 152

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,9 +38,6 @@
     "autoload-dev": {
         "classmap": ["tests/src"]
     },
-    "conflict": {
-        "phpstan/phpstan": "^0.12.43"
-    },
     "extra": {
         "branch-alias": {
             "dev-master": "0.12-dev"
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     "require": {
         "php": "^7.1",
         "nette/finder": "^2.5",
-        "phpstan/phpstan": "dev-master",
+        "phpstan/phpstan": "^0.12.64",
         "symfony/yaml": "~3.4.5|^4.2",
         "webflo/drupal-finder": "^1.2"
     },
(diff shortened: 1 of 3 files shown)


mglaman:
Here is the problem: `\PHPStan\Reflection\BetterReflection\SourceLocator\AutoloadSourceLocator::findReflection`

```
            foreach ($this->classNodes[$identifierName] as $classNode) {
                if ($startLine !== null) {
                    if (\count($classNode->getNode()->attrGroups) > 0 && \PHP_VERSION_ID < 80000) {
                        $startLine--;
                    }
                    if ($startLine !== $classNode->getNode()->getStartLine()) {
                        continue;
                    }
                }
```

The logic fail is on `($startLine !== $classNode->getNode()->getStartLine())`. Due to this reason:

These are lines 10-17 on `PhpunitCompatibilityTrait.php`

```
if (!trait_exists(PhpunitVersionDependentTestCompatibilityTrait::class, FALSE)) {
  class_alias("Drupal\TestTools\PhpUnitCompatibility\PhpUnit" . RunnerVersion::getMajor() . "\TestCompatibilityTrait", PhpunitVersionDependentTestCompatibilityTrait::class);
}

/**
 * Makes Drupal's test API forward compatible with multiple versions of PHPUnit.
 */
trait PhpunitCompatibilityTrait {
```

The reflection from PhpParser for the trait says the startLine is 10 – which is when the `if` statement is defined. But the autoload source locator is passed in the trait start line of 17. So there is a mismatch and it just quits.

mglaman:
If we could have `getNode` check if we can find the identifier and use the startline from the identifier not whatever is the start line of code.

PR #151, opened 2020-11-27
fix: use drupal/recommended-project instead of drupal-composer/drupal…

chrfritsch:
…-project


mglaman:
Thanks, we'll also want to keep D8 pipelines to make sure it works for D8 and D9.

mglaman:
This was fixed in #157

PR #158, opened 2020-12-17
Add 'translation' and 'batch' to hook includes

SpadXIII:
Locale module (and some other contrib modules, like webform) use '.batch.inc' and '.translation.inc' for some code.
Added these to the list of $magic_hook_info_includes so these files are loaded as well.

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -125,6 +125,8 @@ class DrupalAutoloader
                 'tokens',
                 'search_api',
                 'pathauto',
+                'translation',
+                'batch',
             ];
             foreach ($magic_hook_info_includes as $hook_info_include) {
                 if (file_exists($module_dir . "/$module_name.$hook_info_include.inc")) {


mglaman:
So it looks like `locale.module` only has `locale.translation.inc` which is loads in `hook_cron`.

We have `\PHPStan\Rules\Drupal\LoadIncludes` but it looks like that only handles `loadInclude` for the module handler. We need something similar for module_load_include.

src/Drupal/DrupalAutoloader.php
```diff
@@ -125,6 +125,8 @@ public function register(Container $container): void
                 'tokens',
                 'search_api',
                 'pathauto',
+                'translation',
+                'batch',
```
mglaman, line 129: Drupal core doesn't define this in system_hook_info. It looks like it's more `locale` module specific. We need to add dynamic load support for module_load_include

mglaman (changes requested):

mglaman:
This will be fixed in #174. It's a more dynamic approach to solve other instances of one-off .inc in modules.

PR #160, opened 2021-02-01
Use PHPUnit\Framework\Test interface instead of TestCase class

SpadXIII:
I recently started getting errors when running phpstan on my drupal code:

> Internal error: Internal error: PHPUnit\Framework\TestCase is not an interface in file <filename.php>
> Run PHPStan with --debug option and post the stack trace to: https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md

This traced down to \PHPStan\Rules\Drupal\GlobalDrupalDependencyInjectionRule:48 where a list of interface names are checked. Though, the first item of this list is an abstract class and not an interface.

### DIFF
--- a/src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
+++ b/src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
@@ -34,7 +34,7 @@ class GlobalDrupalDependencyInjectionRule implements Rule
 
         $whitelist = [
             // Ignore tests.
-            'PHPUnit\Framework\TestCase',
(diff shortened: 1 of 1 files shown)


damienmckenna:
This fixed the problem for me, thank you.

MPParsley:
+1
```
interface Test extends Countable
abstract class TestCase extends Assert implements SelfDescribing, Test
```

MPParsley (approved):

pobtastic:
+1

mglaman:
Thank you for providing a PR and fix!

Issue #150, opened 2020-11-27
Add PHP8 support


chrfritsch:
It's currently not possible to install this package on PHP8.

`
  [InvalidArgumentException]                                                                                                    
  Package mglaman/phpstan-drupal has a PHP requirement incompatible with your PHP version, PHP extensions and Composer version  
`

mglaman:
This is probably blocked on https://github.com/mglaman/phpstan-drupal/issues/152

wallon-ines:
it's work for me in PHP8 if add _is_string()_  in src/Drupal/DrupalAutoloader.php
```
                            if (is_string($argument)) {
                                $argument = str_replace('@', '', $argument);
                            }
```

klausi:
I have disabled PHPStan testing on PHP 8 only for graphql in the meantime like so: https://github.com/drupal-graphql/graphql/blob/8.x-4.x/.github/workflows/testing.yml#L24

Now that the PR from above is merged maybe just a new release is missing? Or do you want to finish up your Drupal 9 testing first?

vladivo:
@klausi as a workaround until next release you can require master branch:

```bash
composer require --dev --ignore-platform-req=php mglaman/phpstan-drupal:dev-master
```

andypost:
The commited fix still not solving the issue, so I filed follow-up https://github.com/mglaman/phpstan-drupal/pull/163 

Basically container's YML-parser can return int/null/boolean but this helper only supposed to fix service names - strings that starts with `@`

Fixing it this way allowed to run upgrade_status on latest core and PHP 8

mglaman:
I believe we can close this :)

PR #161, opened 2021-02-08
fix TypeError in DrupalAutoloader

wallon-ines:
fix : 
```
TypeError thrown in /project/vendor/mglaman/phpstan-drupal/src/Drupal/DrupalAutoloader.php on line 183 while loading bootstrap file /project/vendor/mglaman/phpstan-drupal/drupal-autoloader.php: str_replace(): Argument #3 ($subject) must be of type array|string, bool given
```
#150

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -175,7 +175,7 @@ class DrupalAutoloader
                 // and thinking these are real services for PHPStan's container.
                 if (isset($serviceDefinition['arguments']) && is_array($serviceDefinition['arguments'])) {
                     array_walk($serviceDefinition['arguments'], function (&$argument) : void {
-                        if (is_array($argument)) {
+                        if (is_array($argument) || !is_string($argument)) {
                             // @todo fix for @http_kernel.controller.argument_metadata_factory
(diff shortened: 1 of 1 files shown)


mglaman (approved):

vladivo:
@mglaman I also ran into this issue while setting up Drupal 9.1.4 with PHP 8 and made my own fix before I found this one. Basically there are several service arguments that are arrays, strings may need replacement and all scalar values (`integer`, `boolean`, `NULL`) do not need any processing. So, I wrote a recursive method, replaced the whole 'array_walk'-block and successfully ran phpstan against Drupal core. To be honest, I am not absolutely sure what exactly I am doing there, but it works and also seems to be formally correct. Have a look, I can create a PR if its okay and allow PHP 8 in requirements. It also resolves the TODO for `@http_kernel.controller.argument_metadata_factory`. Cheers.

```php
private function adjustServiceArguments(array &$arguments) {  
    foreach ($arguments as &$argument) {    
        if (is_array($argument)) {    
            $this->adjustServiceArguments($argument);    
        } elseif (is_string($argument)) {    
            $argument = str_replace('@', '', (string) $argument);    
        }    
    }    
}    
```

PR #167, opened 2021-03-15
Don't try to check interfaces that don't exist

prudloff-insite:
If phpunit is not installed, `GlobalDrupalDependencyInjectionRule` will crash:
```
         Internal error: Interface PHPUnit\Framework\Test does not exist       
         Run PHPStan with --debug option and post the stack trace to:          
         https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md  
```
This PR makes sure the interface exists before calling `implementsInterface()`.

### DIFF
--- a/src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
+++ b/src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
@@ -47,7 +47,7 @@ class GlobalDrupalDependencyInjectionRule implements Rule
         ];
         $classReflection = $scopeClassReflection->getNativeReflection();
         foreach ($whitelist as $item) {
-            if ($classReflection->implementsInterface($item)) {
+            if (interface_exists($item) && $classReflection->implementsInterface($item)) {
                 return [];
             }
         }


mglaman:
Ah yeah, that makes sense! I'm going to see if I can get a test workflow added which runs without `drupal/core-dev` to verify this.

src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
```diff
@@ -47,7 +47,7 @@ public function processNode(Node $node, Scope $scope): array
         ];
         $classReflection = $scopeClassReflection->getNativeReflection();
         foreach ($whitelist as $item) {
-            if ($classReflection->implementsInterface($item)) {
+            if (interface_exists($item) && $classReflection->implementsInterface($item)) {
```
mglaman, line 50: I think we can work around this by getting the interfaces names versus checking if it implements the interface. Working on a new PR.

mglaman:
We still receive the following, but PHPStan does not crash:

```
 ------ -------------------------------------------------------------------- 
  Line   tests/src/Functional/DynamicPageCacheIntegrationTest.php            
 ------ -------------------------------------------------------------------- 
         Class PHPUnit\Framework\TestCase not found.                         
  8      Class PHPUnit\Framework\TestCase not found.                         
  23     Class PHPUnit\Framework\TestCase not found.                         
 ------ -------------------------------------------------------------------- 
```

@ prudloff-insite ☝️  this is should be much better than a full crash! Thanks for reporting the bug.

mglaman:
Fixing in https://github.com/mglaman/phpstan-drupal/pull/167

PR #178, opened 2021-05-27
issue #142: resolve incorrect $reflection definition

### LINKED ISSUE #142: Incorrect inspection of a yaml plugin manager
**Environment:**
 - OS: Linux
 - PHP Version: 7.4.8
 - Drupal core: 8.9.6

**Describe the bug**
With a YAML plugin manager, the `__construct` method has to be overridden, but running `drupal-check` produces an error even though the `__construct` exists in the class and does not call the parent `__construct`. 

```
composer create-project drupal/recommended-project:^8.9.2 drupal8
cd drupal8
composer require mglaman/drupal-check phpstan/phpstan:0.12.42 drupal/console drush/drush
vendor/bin/drush si standard --db-url=mysql://[dbuser]:[dbpass]@[location]/[database] -y
vendor/bin/drupal gm --module=test_module --machine-name=test_module -n
vendor/bin/drupal generate:plugin:type:yaml --module=test_module --class=PluginManager -n
vendor/bin/drupal-check web/modules/custom
```

When I looked though the error I can see that https://github.com/mglaman/phpstan-drupal/blob/master/src/Rules/Classes/PluginManagerInspectionRule.php#L109  has `PhpParser\Node\Name` as the `$constructor->class` which doesn't match the class being inspected `Drupal\test_module\PluginManagerManager`

**Console output**

```
$ vendor/bin/drupal-check web/modules/custom/ -vvv
Current working directory: /var/www/drupal8
Using Drupal root: /var/www/drupal8/web
Using vendor root: /var/www/drupal8/vendor
PHPStan configuration:
parameters:
	tipsOfTheDay: false
	reportUnmatchedIgnoredErrors: false
	excludes_analyse:
		- */tests/Drupal/Tests/Listeners/Legacy/*
		- */tests/fixtures/*.php
		- */settings*.php
		- */node_modules/*

	drupal:
		drupal_root: /var/www/drupal8/web

	customRulesetUsed: true
	ignoreErrors:
		- "#\\Drupal calls should be avoided in classes, use dependency injection instead#"
		- "#Plugin definitions cannot be altered.#"
		- "#Missing cache backend declaration for performance.#"
		- "#Plugin manager has cache backend specified but does not declare cache tags.#"

	bootstrapFiles:
		- /var/www/drupal8/vendor/mglaman/drupal-check/src/Command/../../error-bootstrap.php

includes:
	- /var/www/drupal8/vendor/mglaman/drupal-check/src/Command/../../../../phpstan/phpstan-deprecation-rules/rules.neon
	- /var/www/drupal8/vendor/mglaman/drupal-check/src/Command/../../../../mglaman/phpstan-drupal/extension.neon

Result cache not used because the metadata do not match.
 2/2 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100% < 1 sec/< 1 sec 76.5 MiB

 ------ ----------------------------------------------------------------------- 
  Line   test_module/src/PluginManagerManager.php                               
 ------ ----------------------------------------------------------------------- 
  15     Drupal\test_module\PluginManagerManager must override __construct if   
         using YAML plugins.                                                    
  31      cache tag might be unclear and does not contain the cache key in it.  
 ------ ----------------------------------------------------------------------- 

 [ERROR] Found 2 errors
```

eugene-brit:
Faced with the same problem and after some debug I figure out that `new \ReflectionClass($fqn);` return an incorrect class object.

Fixes #142

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -39,7 +39,6 @@ parametersSchema:
 		))
 	])
 rules:
-	- mglaman\PHPStanDrupal\Rules\Classes\PluginManagerInspectionRule
 	- mglaman\PHPStanDrupal\Rules\Drupal\Coder\DiscouragedFunctionsRule
 	- mglaman\PHPStanDrupal\Rules\Drupal\GlobalDrupalDependencyInjectionRule
 	- mglaman\PHPStanDrupal\Rules\Drupal\PluginManager\PluginManagerSetsCacheBackendRule
@@ -81,6 +80,11 @@ services:
 		tags: [phpstan.broker.propertiesClassReflectionExtension]
 	-
 		class: mglaman\PHPStanDrupal\Rules\Classes\ClassExtendsInternalClassRule
+		tags: [phpstan.rules.rule]
+		arguments:
+			reflectionProvider: @reflectionProvider
+	-
+		class: mglaman\PHPStanDrupal\Rules\Classes\PluginManagerInspectionRule
 		tags: [phpstan.rules.rule]
 		arguments:
 			reflectionProvider: @reflectionProvider
--- a/src/Rules/Classes/PluginManagerInspectionRule.php
+++ b/src/Rules/Classes/PluginManagerInspectionRule.php
@@ -4,11 +4,23 @@ namespace mglaman\PHPStanDrupal\Rules\Classes;
 
 use PhpParser\Node;
 use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
 use PHPStan\Rules\Rule;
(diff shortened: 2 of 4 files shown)


mglaman (approved):
Oof. Thanks for catching this.

mglaman (changes requested):
Looks like we'll need some kind of type hinting somewhere, I guess.

```
  102    Parameter #1 $argument of class ReflectionClass constructor expects  
         class-string<T of object>|T of object, string given.     
```

PR #180, opened 2021-07-02
Use excludePath instead of excludes_analyse

### LINKED ISSUE #179: Switch from excludes_analyse to excludePaths and require phpstan 0.12.65?
Hey,

Huge thanks for this! If I use this project and the extension installer with a local phpstan config that uses [`excludePaths`](https://github.com/phpstan/phpstan/releases/tag/0.12.65), eg.
```neon
parameters:
  excludePaths:
    - path/to/ignore
```
I get the error
```
Configuration parameters excludes_analyse and excludePaths cannot be used at the same time.
```
I can use `excludes_analyse` fine, so it's not a blocker, but I was wondering if it makes sense to require 0.12.65 and switch to the new format?

Thanks!

andriokha:
The two can't be used together and `excludes_analyse` was deprecated in phpstan 0.12.65. See:
- https://phpstan.org/user-guide/ignoring-errors#excluding-whole-files
- https://newreleases.io/project/github/phpstan/phpstan/release/0.12.65
- https://github.com/phpstan/phpstan/issues/3749#issuecomment-754749316

Fixes #179

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ To exclude tests from analysis, add the following parameter
 
 ```
 parameters:
-	excludes_analyse:
+	excludePaths:
 		- *Test.php
 		- *TestBase.php
 ```
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "nette/finder": "^2.5",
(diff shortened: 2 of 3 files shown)


mglaman (approved):
Looks great to me

mglaman:
Same fails as HEAD. I'll fix that up and tag a new release.

PR #181, opened 2021-07-05
fix #147 Magic properties for FieldItemListInterface not detected

### LINKED ISSUE #147: Magic properties for FieldItemListInterface not detected
In my code I have something that boils down to the following statement:

```php
/** @var \Drupal\node\NodeInterface $node */
$node->get('field_name')->value;
```

This results in the following error: `Access to an undefined property Drupal\Core\Field\FieldItemListInterface::$value`.

I find debugging this quite difficult but when I dig into it it seems as if the problem is caused by the usage of `$reflection->implementsInterface()` in [`EntityFieldsViaMagicReflectionExtension`](https://github.com/mglaman/phpstan-drupal/blob/master/src/Reflection/EntityFieldsViaMagicReflectionExtension.php). In my tests this returns `false` for instances of the actual object type like what is returned by [`EntityFieldReflection`](https://github.com/mglaman/phpstan-drupal/blob/master/src/Reflection/EntityFieldReflection.php#L44).

Unfortunately I cannot post the full source code. I would also expect the problem to be picked up by the current testing but it is not.

I am running the analysis using Drupal Check installed as a project dependency using Composer with the following versions:

```
mglaman/drupal-check                            1.1.3
mglaman/phpstan-drupal                          0.12.6
phpstan/phpstan                                 0.12.42
```

When I analyze the same code with a standalone version of Drupal Check using dependencies in their locked versions the error does not occur. Downgrading `phpstan/phpstan` to 0.12.32 also makes the problem go away. Perhaps the situation is caused by a problem upstream.

wallon-ines:
Fix  Magic properties for FieldItemListInterface not detected #147

### DIFF
--- a/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
+++ b/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
@@ -26,7 +26,7 @@ class EntityFieldsViaMagicReflectionExtension implements PropertiesClassReflecti
             // Content entities have magical __get... so it is kind of true.
             return true;
         }
-        if ($reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface')) {
+        if ($reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface') || $reflection->implementsInterface('Drupal\Core\TypedData\ListInterface')) {
             return FieldItemListPropertyReflection::canHandleProperty($classReflection, $propertyName);
         }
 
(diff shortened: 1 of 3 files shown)


mglaman:
Can we have a test? I'm concerned here as field item list implements that interface

wallon-ines:
> Can we have a test? I'm concerned here as field item list implements that interface

I just renamed tests/src/EntityTestClass.php in tests/src/EntityTest.php

mglaman:
I think #244 fixed this. And the test class is outdated. I don't think the existing test coverage is perfect, and the existing test fixture has the same `@todo` comment, but I don't think it's being tested anymore. I've opened https://github.com/mglaman/phpstan-drupal/issues/723 to address this, which will hopefully uncover if https://github.com/mglaman/phpstan-drupal/issues/147 still exists.

PR #195, opened 2021-07-21
Ensure anonymous classes are not reflected

### LINKED ISSUE #194: Anonymous classes trigger errors with version 0.12.11
Since the release of version 0.12.11 we're now seeing errors that look like: 

```
  ------ ------------------------------------------------------------------------------ 
  Line   modules/contrib/localgov_directories/src/Plugin/Block/ChannelSearchBlock.php  
 ------ ------------------------------------------------------------------------------ 
  87     Class                                                                         
         Drupal\localgov_directories\Plugin\Block\AnonymousClass01a859945ded72         
         cee689e1589045065e was not found while trying to analyse it -                 
         discovering symbols is probably not configured properly.                      
 ------ ------------------------------------------------------------------------------ 
```

An example of the code that triggers this error can be seen here: https://github.com/localgovdrupal/localgov_directories/blob/7877f4c15e164b8524df30a0db343865c3621d5e/src/Plugin/Block/ChannelSearchBlock.php#L87

We don't get these errors when pinning phpstan-drupal to version 0.12.10. We're using phpstan 0.12.92.

Are there changes in Drupal coding standards for this or is it an error?

mglaman:
Fixes #194﻿

### DIFF
--- a/src/Rules/Deprecations/DeprecatedAnnotationsRuleBase.php
+++ b/src/Rules/Deprecations/DeprecatedAnnotationsRuleBase.php
@@ -43,6 +43,9 @@ abstract class DeprecatedAnnotationsRuleBase implements Rule
         if ($node->name === null) {
             return [];
         }
+//        if ($node->isAnonymous() || $node->isAbstract()) {
+//            return [];
+//        }
         $className = $node->name->name;
         $namespace = $scope->getNamespace();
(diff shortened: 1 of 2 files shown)


mglaman:
So `isAnonymous` doesn't work because PHPStand gives it a name. See https://github.com/nikic/PHP-Parser/issues/554

Issue #215, opened 2021-10-18
TestCase not found with BrowserTestBaseDefaultThemeRule


codebymikey:
The `BrowserTestBaseDefaultThemeRule` class throws the false positive error:
```
Class PHPUnit\Framework\TestCase not found.
```
when scanning a normal subclass e.g. Forms, even though it doesn't necessarily depend on PHPUnit itself.

Related to #167

codebymikey:
That makes a lot of sense actually, I've added a similar check into the Rule so that it only does the check when PHPUnit is actually available in the class loader. The alternative is to add a try catch instead for the `ClassAutoloadingException`.

Otherwise the actual crash occurs from within [PHPStan's reflection class](https://github.com/phpstan/phpstan-src/blob/629ccf6b02146e1cbf0499982358c595e3d6edd5/src/Reflection/Runtime/RuntimeReflectionProvider.php#L196):

```
#0 (): PHPStan\Reflection\Runtime\RuntimeReflectionProvider->PHPStan\Reflection\Runtime\{closure}('PHPUnit\Framework\TestCase')
#1 /app/application/web/core/tests/Drupal/Tests/BrowserTestBase.php(40): spl_autoload_call('PHPUnit\Framework\TestCase')
#2 phar:///app/application/vendor/phpstan/phpstan/phpstan.phar/vendor/composer/ClassLoader.php(480): include('/app/application/web/core/tests/Drupal/Tests/BrowserTestBase.php')
#3 phar:///app/application/vendor/phpstan/phpstan/phpstan.phar/vendor/composer/ClassLoader.php(346): Composer\Autoload\includeFile('/app/application/web/core/tests/Drupal/Tests/BrowserTestBase.php')
#4 (): Composer\Autoload\ClassLoader->loadClass('Drupal\Tests\BrowserTestBase')
#5 (): spl_autoload_call('Drupal\Tests\BrowserTestBase')
#6 phar:///app/application/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/Runtime/RuntimeReflectionProvider.php(147): class_exists('Drupal\Tests\BrowserTestBase')
#7 phar:///app/application/vendor/phpstan/phpstan/phpstan.phar/src/Reflection/ReflectionProvider/ClassBlacklistReflectionProvider.php(40): PHPStan\Reflection\Runtime\RuntimeReflectionProvider->hasClass('Drupal\Tests\BrowserTestBase')
[...]

PR #216, opened 2021-10-20
Avoid false positives when PHPUnit doesn't exist.

codebymikey:
Addresses #215


mglaman:
I added this, which should fix the root problem 

```
        // @todo stop requiring the bootstrap.php and just copy what is needed.
        if (interface_exists(\PHPUnit\Framework\Test::class)) {
            require $this->drupalRoot . '/core/tests/bootstrap.php';
        }
```

Can you check if that fixes your error?

mglaman:
This is a much larger problem. This does not fix it, see https://github.com/mglaman/phpstan-drupal/pull/246

mglaman:
@codebymikey sorry for closing this, it is an important part of the fix, Sorry. I'll be merging this shortly.

mglaman (approved):

PR #219, opened 2021-11-03
Fix remaining test and analysis errors

mglaman:
Fixes #222

PR #220, opened 2021-11-03
Resolve return of ClassResolver::getInstanceFromDefinition

brambaud:
## Context

`ClassResolver::getInstanceFromDefinition` return type is not currently resolved due to its dynamic nature.

## Observed behaviour 

Given the code:

```php
class SomeClass {
  public function someMethod(): void {}
}

\Drupal::service('class_resolver')
    ->getInstanceFromDefinition(SomeClass::class)
    ->someMethod();
  \Drupal::service('class_resolver')
    ->getInstanceFromDefinition('the_service_id')
    ->someMethod();
```

The following errors are produced:

```
Call to an undefined method object::someMethod().
Call to an undefined method object::anotherMethod().
```

## Expected behaviour

Type should be resolved dynamically and no error should be reported.

Quick question for the tests that are currently missing in this PR, do you have a preferred way to test it to be consistent with the rest of the project?
It seems that neither `DrupalServiceDynamicReturnTypeExtension` nor `ContainerDynamicReturnTypeExtension` are directly tested (maybe I missed them).


src/Type/DrupalClassResolverDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public function getTypeFromMethodCall(
+        MethodReflection $methodReflection,
+        MethodCall $methodCall,
+        Scope $scope
+    ): Type {
+        if (0 === \count($methodCall->args)) {
+            return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+        }
```
mglaman, line 47: See https://github.com/phpstan/phpstan/releases/tag/1.0.0, we need to call `->getArgs()`

mglaman (changes requested):
One nit for the PHPStan 1.0 compatibility

mglaman:
@brambaud you're right about the test. they're indirectly tested and I'm trying to fix that in #219

src/Type/DrupalClassResolverDynamicReturnTypeExtension.php
brambaud: Done
Missed that one thanks!

brambaud:
> @brambaud you're right about the test. they're indirectly tested and I'm trying to fix that in #219

Nice! I'm going to check to add the missing test here then!
Since the part about the service map and retrieving the type from it is similar to `DrupalServiceDynamicReturnTypeExtension` and `ContainerDynamicReturnTypeExtension`, do you already have something in mind? In order to stay consistent with those parts in this PR :)

mglaman:
I do not, entirely. Everything was tested by inference before - could a line of code be scanned and the next line return an expected error (or not one.)

That PR I referenced also has a wild failure I need to resolve.

There is a test porter for a different reflection extension, maybe copy that? Or how PHPStan itself handles those. I think I may have looked at Larastan or phpstan-symfony for inspiration. On mobile, so hard to find references right now

brambaud:
I have added the test :)

I had to create the directories `tests/fixtures/drupal/profiles` and `tests/fixtures/drupal/themes` because they are required by the drupal test bootstrap (`tests/fixtures/drupal/core/tests/bootstrap.php`).

I choose to create a module with an available service that could be reused for `DrupalServiceDynamicReturnTypeExtension` and `ContainerDynamicReturnTypeExtension`.

Since `DrupalServiceDynamicReturnTypeExtension` and `ContainerDynamicReturnTypeExtension` are similar, do you want that I add tests for them too here or do you prefer in another PR (maybe after merging this one to reuse what have been started here 🤔 )?

Also, the part getting the type from the service definition is the same between those 3 classes:

```php
private function getTypeFromService(DrupalServiceDefinition $serviceDefinition, string $serviceId): Type
    {
        // Work around Drupal misusing the SplString class for string
        // pseudo-services such as 'app.root'.
        // @see https://www.drupal.org/project/drupal/issues/3074585
        if ($serviceDefinition->getClass() === 'SplString') {
            return new StringType();
        }

        return new ObjectType($serviceDefinition->getClass() ?? $serviceId);
    }
```

do you think we should try to reuse this for the three classes?

brambaud:
Ho! Just found out that `\Drupal::classResolver()` exists 😅 

In order to cover this case too, we need to create another DynamicStaticMethodReturnTypeExtension

brambaud:
Done for the support of `\Drupal::classResolver()`.

Nevertheless not really convince by what I've done!

mglaman:
> do you want that I add tests for them too here or do you prefer in another PR (maybe after merging this one to reuse what have been started here 🤔 )?

We can do as a follow up PR

> Also, the part getting the type from the service definition is the same between those 3 classes:

Maybe we could make a shared based class that has a protected method? 

Haven't reviewed the code yet, just replying to comments. I'll swing by the code today for a review

src/Type/DrupalClassResolverReturnType.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private static function getTypeFromService(DrupalServiceDefinition $serviceDefinition, string $serviceId): Type
+    {
+        // Work around Drupal misusing the SplString class for string
+        // pseudo-services such as 'app.root'.
+        // @see https://www.drupal.org/project/drupal/issues/3074585
+        if ($serviceDefinition->getClass() === 'SplString') {
+            return new StringType();
+        }
```
mglaman, line 47: Actually, this could be a static method on DrupalServicedefinition, maybe? 

We could provide a `getType` on the definition class. It breaks the "copy and paste" from Drupal core. But whatever, it's forked code anyways. Then we do not need workarounds as the workaround logic is now encapsulated in the class itself.

brambaud: Nice idea to add it to DrupalServiceDefinition!

I choose a non-static method since right now everywhere it is needed we have the DrupalServiceDefinition directly.

mglaman (approved):
Merge in fixes from `main` to get tests green!

mglaman:
THANK YOU @brambaud. I am beyond excited to see this merge in, thank you for improving the static analysis capabilities!

PR #227, opened 2021-11-11
Test drupal container related type inference

brambaud:
Follow up of #220

### DIFF
--- a/tests/src/Type/DrupalContainerDynamicReturnTypeTest.php
+++ b/tests/src/Type/DrupalContainerDynamicReturnTypeTest.php
@@ -7,12 +7,14 @@ namespace mglaman\PHPStanDrupal\Tests\Type;
 use mglaman\PHPStanDrupal\Tests\AdditionalConfigFilesTrait;
 use PHPStan\Testing\TypeInferenceTestCase;
 
-class DrupalClassResolverDynamicReturnTypeExtensionTest extends TypeInferenceTestCase
+final class DrupalContainerDynamicReturnTypeTest extends TypeInferenceTestCase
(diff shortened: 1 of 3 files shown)


tests/src/Type/DrupalContainerRelatedTypeInferenceTest.php
```diff
@@ -7,12 +7,14 @@
 use mglaman\PHPStanDrupal\Tests\AdditionalConfigFilesTrait;
 use PHPStan\Testing\TypeInferenceTestCase;
 
-class DrupalClassResolverDynamicReturnTypeExtensionTest extends TypeInferenceTestCase
+final class DrupalContainerRelatedTypeInferenceTest extends TypeInferenceTestCase
```
brambaud, line 10: Not sure about the naming. Is it understandable enough? 🤔

mglaman (commented):
I appreciate this follow up! @brambaud these have been some of the best contributions from someone so far. It greatly improved the static analysis capabilities.

tests/src/Type/DrupalContainerRelatedTypeInferenceTest.php
brambaud: Yep that is exactly that!

brambaud:
Thanks!
If you need some help on specific tasks let me know if there are some with higher priority than others!

mglaman (approved):

mglaman:
@brambaud will do! I've started opening more issues as folks try to run this against Drupal core for analysis. If anything looks interesting, feel free. There isn't an exact plan

Issue #238, opened 2021-11-15
Entity class return type extension from entity storage methods


mglaman:
We have a return type extension to identify the storage class from `getStorage`. It'd be amazing if we could get methods from the storage to return the correct entity:

* `create`
* `load`
* `loadMultiple`
* `loadUnchanged`
* `loadByProperties`

PR #239, opened 2021-11-16
Support entity class return type extension from entity storage methods

brambaud:
Closing #238

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -17,10 +17,17 @@ parameters:
 			node: Drupal\node\NodeStorage
 			taxonomy_term: Drupal\taxonomy\TermStorage
 			user: Drupal\user\UserStorage
+			block: Drupal\Core\Config\Entity\ConfigEntityStorage
+		entityStorageMapping:
+			node: Drupal\node\Entity\Node
+			taxonomy_term: Drupal\taxonomy\Entity\Term
+			user: Drupal\user\Entity\User
+			block: Drupal\block\Entity\Block
 parametersSchema:
 	drupal: structure([
 		drupal_root: string()
 		entityTypeStorageMapping: arrayOf(string())
+		entityStorageMapping: arrayOf(string())
 	])
 rules:
 	- mglaman\PHPStanDrupal\Rules\Classes\PluginManagerInspectionRule
@@ -35,7 +42,13 @@ services:
 	-
 		class: mglaman\PHPStanDrupal\Type\EntityTypeManagerGetStorageDynamicReturnTypeExtension
 		arguments:
+			reflectionProvider: @reflectionProvider
 			entityTypeStorageMapping: %drupal.entityTypeStorageMapping%
+		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+	-
+		class: mglaman\PHPStanDrupal\Type\EntityStorage\EntityStorageDynamicReturnTypeExtension
+		arguments:
+			entityStorageMapping: %drupal.entityStorageMapping%
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 	-
 		class: mglaman\PHPStanDrupal\Type\ContainerDynamicReturnTypeExtension
--- a/src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
+++ b/src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
@@ -2,6 +2,11 @@
 
 namespace mglaman\PHPStanDrupal\Type;
 
+use Drupal\Core\Config\Entity\ConfigEntityStorageInterface;
+use Drupal\Core\Entity\ContentEntityStorageInterface;
+use mglaman\PHPStanDrupal\Type\EntityStorage\ConfigEntityStorageType;
+use mglaman\PHPStanDrupal\Type\EntityStorage\ContentEntityStorageType;
+use mglaman\PHPStanDrupal\Type\EntityStorage\EntityStorageType;
 use PhpParser\Node\Expr\BinaryOp\Concat;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Scalar\String_;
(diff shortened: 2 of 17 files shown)


mglaman:
Thanks @brambaud !!! I will give this a more in-depth look over. I love the tests.

src/Type/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,94 @@
...
+        if (!$callerType instanceof ObjectType) {
+            return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+        }
+
+        $entityTypeStorageMapping = \array_flip($this->entityTypeStorageMapping);
+        $entityType = $entityTypeStorageMapping[$callerType->getClassName()] ?? null;
+        if (null === $entityType) {
+            return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+        }
+
```
mglaman, line 77: I don't think we need to check if the entity storage was mapped, at all. If we don't have the mapping, it's perfectly fine for us to trust the default storage interface BUT define the entity type class returned.

mglaman (changes requested):
I'd like to drop the check to see if the entity type storage mapping exists. Not all entities have a custom storage that adds methods, so falling back on the default interface is okay.

src/Type/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -0,0 +1,94 @@
...
+        $type = new ObjectType($entityClassName);
+        if (\in_array($methodReflection->getName(), ['load', 'loadUnchanged'], true)) {
+            return TypeCombinator::addNull($type);
+        }
+
+        if (\in_array($methodReflection->getName(), ['loadMultiple', 'loadByProperties'], true)) {
+            return new ArrayType(new IntegerType(), $type);
+        }
```
mglaman, line 90: We need to check the type of storage. If it is ConfigEntityStorageInterface it is `array<string, $type>`.

mglaman (changes requested):

src/Type/EntityStorageDynamicReturnTypeExtension.php
mglaman: We should add mapping for Block config and test that here.

brambaud: Done! for both!

src/Type/EntityStorageDynamicReturnTypeExtension.php
brambaud: This part has been removed.

src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
```diff
@@ -2,19 +2,30 @@
 
 namespace mglaman\PHPStanDrupal\Type;
```
brambaud, line 3: I did not change the namespace of this class to avoid BC break in case someone already extend it.

What do you think about: 
- copy/paste it to `mglaman\PHPStanDrupal\Type\EntityStorage` namespace
- make it extends the new class in `mglaman\PHPStanDrupal\Type\EntityStorage` namespace
- deprecate it

or do you think it's ok to do the BC break?

src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
```diff
@@ -23,10 +34,12 @@ class EntityTypeManagerGetStorageDynamicReturnTypeExtension implements DynamicMe
     /**
      * EntityTypeManagerGetStorageDynamicReturnTypeExtension constructor.
      *
+     * @param ReflectionProvider $reflectionProvider
      * @param string[] $entityTypeStorageMapping
      */
-    public function __construct(array $entityTypeStorageMapping = [])
+    public function __construct(ReflectionProvider $reflectionProvider, array $entityTypeStorageMapping = [])
```
brambaud, line 40: If you like to avoid the BC break I will have to add the reflection provider argument in second

brambaud:
In order to avoid the solution based on the storage mapping, it introduced 3 new types:
- EntityStorageType: to store the entity type id in order to reuse it for the mapping.
- ConfigEntityStorageType: to identify config entity storage and then config entity.
- ContentEntityStorageType: to identify content entity storage and then content entity.

More tests have been added too.
Not sure all cases are covered though 🤔
And I don't really found them readable. Maybe grouping cases using function could allow easier reading and understanding. 🤔 
Need some sleep to think about it 😅

src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
mglaman: I don't care about backward compatibility with this library. The only namespace that I would care around BC are Rules, but none of them are opt-in yet, they just run. A decision I'd like to change.

Break the BC! 😄

src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
mglaman: Nah, no need to avoid.

mglaman (commented):
Regarding my thoughts on the EntityTypeDataRepository, I'd like to know what you think @brambaud. I am tempted to merge this _and then_ introduce the EntityTypeDataRepository. That way this PR doesn't languish any further.

PS: this is amazing work.

mglaman:
Also, I think the test coverage is top-notch. I appreciate taking the time to add some custom entity definitions, as we'll be able to leverage those more later on.

mglaman:
I'm going to merge this, today, and work on follow ups :)

mglaman (approved):
Approving, doing the EntityDataRepository as a follow up

brambaud:
Sorry for the late reply :/

I'm totally ok for the EntityDataRepository! Great idea!

extension.neon
```diff
@@ -17,10 +17,17 @@ parameters:
 			node: Drupal\node\NodeStorage
 			taxonomy_term: Drupal\taxonomy\TermStorage
 			user: Drupal\user\UserStorage
+			block: Drupal\Core\Config\Entity\ConfigEntityStorage
+		entityStorageMapping:
+			node: Drupal\node\Entity\Node
+			taxonomy_term: Drupal\taxonomy\Entity\Term
+			user: Drupal\user\Entity\User
+			block: Drupal\block\Entity\Block
```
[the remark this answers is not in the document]
brambaud: I wanted to propose something like this and I was afraid the BC break could be too much but honestly this is the best decision!

Keeping the entity type id has the key is fine I think. I wonder if I will be able to overwrite it as a user 🤔 
I thought about creating a command to generate this mapping to allow specific project to have their specific mapping without having to maintain it manually. It is pretty easy to find the definitions using `\Drupal::entityTypeManager()->getDefinitions()` and create the config automatically.

PR #241, opened 2021-11-16
Support entity query execute dynamic return type whether it is a count query or not

brambaud:
`\Drupal\Core\Entity\Query\QueryInterface::execute` return type depends whether it is a count query or not.

```
/**
   * Execute the query.
   *
   * @return int|array
   *   Returns an integer for count queries or an array of ids. The values of
   *   the array are always entity ids. The keys will be revision ids if the
   *   entity supports revision and entity ids if not.
   */
  public function execute();
```

For the moment only chaining the methods is supported, i.e

```php
\Drupal::entityTypeManager()->getStorage('node')
        ->getQuery()
        ->accessCheck(TRUE)
        ->execute()
```

 I hope to find a way to support cases like another day:

```php
$query = \Drupal::entityTypeManager()->getStorage('node')
    ->getQuery()
    ->accessCheck(TRUE);
$query->execute();
```


mglaman:
While looking at this, I feel like part of the trick is to also have a dynamic return type on Entity\QueryInterface::count which flags _something_ so we're aware that the parent object type (the query) is a count

mglaman:
@brambaud can you review my push? I ended up rewriting the dynamic return type extension. I added a new custom object type. When we encounter a `count` method, we alter the return type to be this new custom object type. It allows us to check the type and know if it was a count query. Like if the method returned a different interface.

brambaud:
@mglaman I'll try to work on phpstan-drupal at least Friday or this weekend so I'll check that at this moment!
Nice idea for the custom type! I had it too but for #239 😛

brambaud:
Looks great!

It's true `\Drupal::entityQuery` exists too! I'm so used to dependency injection that I forget these static methods exist! 😅 

About the todo, should we wait #239 to be finished and work on this to avoid false positive for config entities?

tests/src/Type/EntityQueryDynamicReturnTypeExtensionTest.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Tests\Type;
```
brambaud, line 5: Shouldn't the test be in a consistent namespace with `mglaman\PHPStanDrupal\Type\EntityQuery\EntityQueryDynamicReturnTypeExtension`?

Something like `mglaman\PHPStanDrupal\Tests\Type\EntityQuery`?

mglaman: Yeah we can move it there

mglaman:
@brambaud I think we can handle the config entity storage as a follow up. Technically it's not a regression either way. So we could merge this or that and then address config entity oddities later

tests/src/Type/EntityQueryDynamicReturnTypeExtensionTest.php
brambaud: Done!

brambaud:
@mglaman Ok! I'll handle the follow up then :)

mglaman (approved):

PR #244, opened 2021-11-19
Allow magic methods for variables of type FieldItemListInterface

eiriksm:
So here is a thing.

If a variable was declared to be just the FieldItemListInterface, phpstan-drupal would not allow it to access things like ->value, since we require it to implement interface FieldItemListInterface. Which returns false if the variable is literally that interface.

You would need at lest level 2 to get this error.

An example:

```
/** @var \Drupal\Core\Field\FieldItemListInterface $field */
$field = $node->get('my_field');
$value = $field->value; // <- this fails with Access to an undefined property Drupal\Core\Field\FieldItemListInterface::$value
```

You don't actually need to typehint the field either, because the ->get method is supposed to return FieldItemListInterface.

```
$field = $node->get('my_field');
$value = $field->value; // <- fails with Access to an undefined property Drupal\Core\Field\FieldItemListInterface::$value
```

So currently you would have to do something like this:

```
/** @var \Drupal\Core\Field\FieldItemList $field */
$field = $node->get('my_field');
$value = $field->value; // <- does not fail.
```

With this PR you can do whichever of the variants you like, they all work :tada:

### DIFF
--- a/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
+++ b/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
@@ -58,7 +58,7 @@ class EntityFieldsViaMagicReflectionExtension implements PropertiesClassReflecti
     {
         $classObject = new ObjectType($reflection->getName());
         $interfaceObject = self::getFieldItemListInterfaceObject();
-        return $classObject->isSuperTypeOf($interfaceObject);
+        return $interfaceObject->isSuperTypeOf($classObject);
     }
 
     protected static function getFieldItemListInterfaceObject() : ObjectType
--- a/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
+++ b/src/Reflection/EntityFieldsViaMagicReflectionExtension.php
@@ -5,6 +5,7 @@ namespace mglaman\PHPStanDrupal\Reflection;
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertiesClassReflectionExtension;
 use PHPStan\Reflection\PropertyReflection;
+use PHPStan\TrinaryLogic;
 use PHPStan\Type\ObjectType;
 
 /**
@@ -33,7 +34,7 @@ class EntityFieldsViaMagicReflectionExtension implements PropertiesClassReflecti
             // Content entities have magical __get... so it is kind of true.
             return true;
         }
-        if (self::classObjectIsSuperOfFieldItemList($reflection)) {
+        if (self::classObjectIsSuperOfFieldItemList($reflection)->yes()) {
             return FieldItemListPropertyReflection::canHandleProperty($classReflection, $propertyName);
         }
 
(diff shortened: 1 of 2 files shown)


eiriksm:
I guess this would close #181 and fix #147 

Sorry for not searching first :(

src/Reflection/EntityFieldsViaMagicReflectionExtension.php
```diff
@@ -32,7 +32,7 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
             // Content entities have magical __get... so it is kind of true.
             return true;
         }
-        if ($reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface')) {
+        if ($reflection->getName() === 'Drupal\Core\Field\FieldItemListInterface' || $reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface')) {
```
mglaman, line 35: I think this is a great time to leverage ObjectType here. We can use an object type for the interface and check if the current item is a superOf the interface. If I recall, being the same returns true. Helps streamline the conditions

src/Reflection/EntityFieldsViaMagicReflectionExtension.php
```diff
@@ -45,7 +45,7 @@ public function getProperty(ClassReflection $classReflection, string $propertyNa
         if ($reflection->implementsInterface('Drupal\Core\Entity\EntityInterface')) {
             return new EntityFieldReflection($classReflection, $propertyName);
         }
-        if ($reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface')) {
+        if ($reflection->getName() === 'Drupal\Core\Field\FieldItemListInterface' || $reflection->implementsInterface('Drupal\Core\Field\FieldItemListInterface')) {
```
mglaman, line 48: See previous comment about using types

tests/src/Reflection/EntityFieldsViaMagicReflectionExtensionTest.php
```diff
@@ -70,6 +70,11 @@ public function dataHasProperty(): \Generator
             'target_id',
             true,
         ];
+        yield 'field item list_interface: value' => [
+            \Drupal\Core\Field\FieldItemListInterface::class,
+            'target_id',
+            true,
+        ];
```
mglaman, line 77: Nit: test title is value but property is target_id

mglaman (commented):
@eiriksm would you want to take a stab at converting the reflection checks into object types? I can later. Working with them in other dynamic return types has made the code a bit easier. It's a newer concept since these return type extensions were created

src/Reflection/EntityFieldsViaMagicReflectionExtension.php
eiriksm: OK, I think I changed it to what you had in mind :)

tests/src/Reflection/EntityFieldsViaMagicReflectionExtensionTest.php
eiriksm: Well, I copy pasted it from the test above it, so there is also the same nit there ;)

Changed them both then

mglaman (approved):
Thanks!

bradjones1:
Does this fix https://github.com/mglaman/phpstan-drupal/issues/147 ?

mglaman:
@bradjones1 it might. I think I left that original issue open for generally more complex support

PR #246, opened 2021-11-19
Fix incompatibility without drupal/core-dev

mglaman:
Let's see if we can nail this down.


mglaman:
I need to figure out why this is happeing:

```
 ------ -------------------------------------------------------------------- 
  Line   src/PageCache/RequestPolicy/DefaultRequestPolicy.php                
 ------ -------------------------------------------------------------------- 
  15     Class PHPUnit\Framework\TestCase not found.                         
 ------ -------------------------------------------------------------------- 
```

Looking at `ClassConstantRule`: https://github.com/phpstan/phpstan-src/blob/447cd102d946a2d9883ac82793195275f54296a9/src/Rules/Classes/ClassConstantRule.php#L98 it's due to fetching `PHPUnit\Framework\TestCase::class` somewhere.

mglaman:
The only occurrence I can find is in `\Drupal\Core\Test\RunTests\TestFileParser`. But that is only used in the `run-tests.sh` script.

mglaman:
Here's where the error can come from [link](https://github.com/phpstan/phpstan-src/search?q=%22Class+%25s+not+found.%22):

* ClassAutoloadingException when thrown
* ExistingClassInInstanceOfRule
* ClassConstantRule

mglaman:
It's probably in here: \PHPStan\Reflection\Runtime\RuntimeReflectionProvider::hasClass

Debugging must occur here to see why the exception is thrown.

```
        try {
            return $this->hasClassCache[$className] = \class_exists($className) || \interface_exists($className) || \trait_exists($className);
        } catch (\PHPStan\Broker\ClassAutoloadingException $e) {
            throw $e;
        } catch (\Throwable $t) {
            throw new \PHPStan\Broker\ClassAutoloadingException($className, $t);
        } finally {
            \spl_autoload_unregister($autoloader);
        }
```

Issue #183, opened 2021-07-09
Detect when a class extends or calls a method that has been flagged as @internal


mglaman:
Drupal doesn't use `final` or `private` accessors often. Usually we opt for `@internal`. We should make a rule for detecting this. 

Usage of `@internal` should be allowed if it's in a shared namespace. Such as `Drupal\Core\*` using an `@internal` class in `\Drupal\Core\*`. But it should error if `Drupal\mymodule` is using it.

Issue #247, opened 2021-11-19
Feature Request: Validate that `.info.yml` contains modules from which classes or functions are used


Kingdutch:
In larger projects it's possible that modules are already enabled and thus that a class exists. However, it's possible that when the module being worked on is installed in a different project that the dependency isn't actually there.

It would be great since PHPStan knows what classes and functions are used and where they come from that it can check based on the namespace or path that the module to which the class or function belongs is in the `.info.yml` file of the module that's using it.

mglaman:
@Kingdutch what level are you running PHPStan at? Because this should "just work" if running at higher levels. If you're only doing deprecation checks (drupal-check, upgrade_status) that isn't even `level: 0`. It is a custom runner.

Kingdutch:
We're running [PHPStan at level 8](https://github.com/goalgorilla/open_social/blob/main/phpstan.neon#L5), although we do have an [extensive baseline](https://github.com/goalgorilla/open_social/blob/main/phpstan-baseline.neon). I didn't see any of the warnings I'd expect in the baseline.

I'm not sure I explained correctly. If the feature is implemented I would, using Open Social as example, [a use of a class from `social_core` in `social_user`](https://github.com/goalgorilla/open_social/blob/main/modules/social_features/social_user/src/Entity/User.php#L5) to emit an error if [`social_core` is not listed as dependency](https://github.com/goalgorilla/open_social/blob/main/modules/social_features/social_user/social_user.info.yml). Such error is not currently part of PHPStan Drupal as far as I could find.

mglaman:
@Kingdutch oh! so try to report that you're accessing namespaces which _may not actually_ be available at runtime.

I think this got up before. Because right now this projects consumes all possible namespaces. So an idea would be to allow configuring what modules should be scanned – either manually or scanning the `core.extensions.yml`.

It sounds like that is what you need. I think it's smarter to try and parse `core.extensions.yml` than to make someone manually track in `phpstan.neon` and equally dangerous to become out of date

Kingdutch:
Not entirely! I think that's what I saw in #4 ?

I don't want to change the PHPStan configuration. What I want PHPStan to tell me is "Hey this module A is using code from this other module B, but the module A doesn't tell Drupal (using A's .info.yml file) that B is a dependency, that's an error".

mglaman:
Ah, now I understand! Detecting if a dependent namespace is not available because its dependency on it is not declared.

I think that might be feasible but also very complex.

[...]

PR #248, opened 2021-11-20
Detect when a class extends a class that has been flagged as @internal

brambaud:
See #183

In this PR, a shared namespace for Drupal starts with `Drupal\project_name`.

It means:
* a class inside `Drupal\Core` can extends an internal class from `Drupal\Core`
* a class inside `Drupal\my_module` can extends an internal class from `Drupal\my_module`
* a class inside `Drupal\my_module` cannot extends an internal class from `Drupal\Core` or `\Drupal\another_module`
* a class inside `Drupal\Component` cannot extends an internal class from `Drupal\Core` or `\Drupal\another_module`
* a class inside `Drupal\Core` cannot extends an internal class from `Drupal\Component` or `\Drupal\another_module`
* etc..

For the part detecting calls to internal methods, I think it can be handled in another PR 🤔

### DIFF
--- a/src/Rules/Classes/ClassExtendsInternalClassRule.php
+++ b/src/Rules/Classes/ClassExtendsInternalClassRule.php
@@ -1,0 +1,75 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Rules\Classes;
+
+use mglaman\PHPStanDrupal\Internal\NamespaceCheck;
+use PhpParser\Node;
+use PhpParser\Node\Stmt\Class_;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+class ClassExtendsInternalClassRule implements Rule
+{
+    /**
+     * @var ReflectionProvider
+     */
+    private $reflectionProvider;
+
+    public function __construct(ReflectionProvider $reflectionProvider)
+    {
+        $this->reflectionProvider = $reflectionProvider;
+    }
+
+    public function getNodeType(): string
+    {
+        return Class_::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        /** @var Class_ $node */
+        if (!isset($node->extends)) {
+            return [];
+        }
+
+        $extendedClassName = $node->extends->toString();
+        if (!$this->reflectionProvider->hasClass($extendedClassName)) {
+            return [];
+        }
+
+        $extendedClassReflection = $this->reflectionProvider->getClass($extendedClassName);
+        if (!$extendedClassReflection->isInternal()) {
+            return [];
+        }
+
+        // @phpstan-ignore-next-line
+        if (!isset($node->namespacedName)) {
+            return $this->buildError(null, $extendedClassName);
+        }
+
+        $currentClassName = $node->namespacedName->toString();
+
+        if (!NamespaceCheck::isDrupalNamespace($node)) {
+            return $this->buildError($currentClassName, $extendedClassName);
+        }
+
+        if (NamespaceCheck::isSharedNamespace($node)) {
+            return [];
+        }
+
+        return $this->buildError($currentClassName, $extendedClassName);
+    }
+
+    private function buildError(?string $currentClassName, string $extendedClassName): array
+    {
+        return [
+            RuleErrorBuilder::message(\sprintf(
+                '%s extends @internal class %s.',
+                $currentClassName !== null ? \sprintf('Class %s', $currentClassName) : 'Anonymous class',
+                $extendedClassName
+            ))->build()
+        ];
+    }
+}
--- a/extension.neon
+++ b/extension.neon
@@ -77,6 +77,11 @@ services:
 		class: mglaman\PHPStanDrupal\Reflection\EntityFieldsViaMagicReflectionExtension
 		tags: [phpstan.broker.propertiesClassReflectionExtension]
 	-
+		class: mglaman\PHPStanDrupal\Rules\Classes\ClassExtendsInternalClassRule
+		tags: [phpstan.rules.rule]
+		arguments:
+			reflectionProvider: @reflectionProvider
+	-
 		class: mglaman\PHPStanDrupal\Rules\Drupal\LoadIncludes
 		tags: [phpstan.rules.rule]
 		arguments:
--- a/src/Internal/NamespaceCheck.php
+++ b/src/Internal/NamespaceCheck.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Internal;
+
+use PhpParser\Node\Stmt\Class_;
+
+/**
+ * @internal
+ */
+final class NamespaceCheck
+{
+    public static function isDrupalNamespace(Class_ $class): bool
+    {
+        // @phpstan-ignore-next-line
+        if (!isset($class->namespacedName)) {
+            return false;
+        }
+
+        return 'Drupal' === (string) $class->namespacedName->slice(0, 1);
+    }
+
+    public static function isSharedNamespace(Class_ $class): bool
+    {
+        if (!isset($class->extends)) {
+            return false;
+        }
+
+        // @phpstan-ignore-next-line
+        if (!isset($class->namespacedName)) {
+            return false;
+        }
+
+        if (!self::isDrupalNamespace($class)) {
+            return false;
+        }
+
+        return (string) $class->namespacedName->slice(0, 2) === (string) $class->extends->slice(0, 2);
+    }
+}
--- a/tests/fixtures/drupal/modules/module_with_internal_classes/module_with_internal_classes.info.yml
+++ b/tests/fixtures/drupal/modules/module_with_internal_classes/module_with_internal_classes.info.yml
@@ -1,0 +1,3 @@
+name: module_with_internal_classes
+type: module
+core: 8.x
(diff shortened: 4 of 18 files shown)


src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+    public function processNode(Node $node, Scope $scope): array
+    {
+        /** @var Class_ $node */
+        if (!isset($node->extends)) {
+            return [];
+        }
+
+        $extendedClassName = $node->extends->toString();
+        if (!$this->reflectionProvider->hasClass($extendedClassName)) {
+            return [];
```
brambaud, line 38: Should we return an error if the extended class is not found?

mglaman: We should not error, because PHPStan has rules for this

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        $extendedClassReflection = $this->reflectionProvider->getClass($extendedClassName);
+        if (!$extendedClassReflection->isInternal()) {
+            return [];
+        }
+
+        if (!isset($node->namespacedName)) {
+            return $this->buildError(null, $extendedClassName);
+        }
```
mglaman, line 48: Should we error here? Some code in PHPStan shows it throws a ShouldNotHappenException 

https://github.com/phpstan/phpstan-src/search?q=namespacedName

mglaman:
This is amazing. I want to come back and review it more. Because we could move a lot of this logic into a shared service that can help handle https://github.com/mglaman/phpstan-drupal/issues/249. 

Like [`FunctionCallParametersCheck`](https://github.com/phpstan/phpstan-src/blob/master/src/Rules/FunctionCallParametersCheck.php) in PHPStan. That namespace service could return errors as needed. To start it'll return errors on the usage of internal across namespaces. Later it can be enhanced to check the current namespace's extension dependency compared to called class.

src/Rules/Classes/ClassExtendsInternalClassRule.php
brambaud: We should not error here because it handles anonymous classes.

PHPStan throws for interface, trait, function etc.. or [when an anonymous class has a namespaced name](https://github.com/phpstan/phpstan-src/blob/629ccf6b02146e1cbf0499982358c595e3d6edd5/src/Reflection/BetterReflection/BetterReflectionProvider.php#L205)

brambaud:
> This is amazing. I want to come back and review it more. Because we could move a lot of this logic into a shared service that can help handle #249.
> 
> Like [`FunctionCallParametersCheck`](https://github.com/phpstan/phpstan-src/blob/master/src/Rules/FunctionCallParametersCheck.php) in PHPStan. That namespace service could return errors as needed. To start it'll return errors on the usage of internal across namespaces. Later it can be enhanced to check the current namespace's extension dependency compared to called class.

I'm not sure how a service like that could be named 🤔 

Do you mean something like:

```php
final class Foo {
  public function isClassFromAnotherExtension(\PhpParser\Node\Stmt\Class_ $node);
  public function IsClassFromADependencyExtension();
}
```

?

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+            ))->build()
+        ];
+    }
+
+    private function isInDrupalNamespace(Class_ $class): bool
+    {
+        return 'Drupal' === (string) $class->namespacedName->slice(0, 1);
+    }
+
+    private function isClassAndExtendedClassInSameDrupalProjectNamespace(Class_ $class): bool
```
mglaman, line 79: I think we can rename this to `isSharedNamespace`. We consider it to be "shared" if the first two levels match:

* \Drupal\Core
* \Drupal\Component
* \Drupal\{extension}

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        if (!isset($class->namespacedName)) {
+            return false;
+        }
```
mglaman, line 83: Isn't this technically handled earlier? We have it as a private method, so it's safe to assume this check.

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+
+        if (!isset($class->extends)) {
+            return false;
+        }
```
mglaman, line 87: This check should go earlier into the Rule

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+            return [];
+        }
+
+        if (!isset($node->namespacedName)) {
+            return $this->buildError(null, $extendedClassName);
+        }
+
+        $currentClassName = $node->namespacedName->toString();
+
+        if (!$this->isInDrupalNamespace($node)) {
```
mglaman, line 52: Let's ditch a standalone method for just `'Drupal' === (string) $class->namespacedName->slice(0, 1)`

src/Rules/Classes/ClassExtendsInternalClassRule.php
```diff
@@ -0,0 +1,91 @@
...
+            return $this->buildError($currentClassName, $extendedClassName);
+        }
+
+        if ($this->isClassAndExtendedClassInSameDrupalProjectNamespace($node)) {
```
mglaman, line 56: We could also ditch the method here for directly checking `(string) $class->namespacedName->slice(0, 2) === (string) $class->extends->slice(0, 2)`

mglaman (commented):
Some thoughts on naming and maybe ditching the private methods added to the Rule.

In reference to #247 (I tagged the wrong issue before), we could move these two methods into static methods in a namespace helper utility class. One that checks if a namespace is in Drupal and another if it's in a shared namespace.

brambaud:
I will try to re-work this PR this week to include the requested changes (hoping I won't have to wait this weekend)

src/Rules/Classes/ClassExtendsInternalClassRule.php
brambaud: I understand the point! I'm used to defensive programming even in private method 😛

brambaud:
ah! Seems the fresh (3h ago) and new [nikic/php-parser version 4.13.2](https://github.com/nikic/PHP-Parser/releases/tag/v4.13.2) breaks the CI thanks to this:

> The namespacedName property populated by the NameResolver is now declared on relevant nodes,
to avoid a dynamic property deprecation warning with PHP 8.2.

I think we could ignore this error because `ClassLike::namespacedName` can still be null even if its true it will always be set.
Furthermore, unless we force the version to be >=4.13.2 it could still break if we only take into account the nullable part.

I'm gonna add this error to the ignored error in the phpstan configuration.

What do you think?

mglaman:
@brambaud could we just add the "ignore next line" inline error? On mobile, but PHPStan has some inline comment to ignore an error on the next line

brambaud:
@mglaman If you prefer the inline ignore tag fine by me!
I'm more used to add the errors directly to the configuration to have only one place to look. The possibility to ignore the same error for all the code source or to comment why it could be ignored is appealing to me.

mglaman:
@brambaud Since it's one failure, I like a single ignore. If we had 30+ and it'd be a huge refactor, then I'd ignore it in the config. Either approach requires coming back to audit errors being ignored. This single ignore helps as a "reminder" if it shows again in new code

mglaman:
I'll give this a review later this evening, I have a feeling it's ready to go :)

src/Internal/ClassHelper.php
```diff
@@ -0,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Internal;
+
+use PhpParser\Node\Stmt\Class_;
+
+/**
+ * @internal
+ */
+final class ClassHelper
```
brambaud, line 10: Not sure about the naming 😅 

I wonder if we could deal with reflection to use a custom type that could hold those data 🤔
Not sure if it worth the effort right now 🤔

src/Internal/ClassHelper.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        // @phpstan-ignore-next-line
+        if (!isset($class->namespacedName)) {
+            return false;
+        }
+
+        return 'Drupal' === (string) $class->namespacedName->slice(0, 1);
+    }
+
+    public static function isSharedNamespace(Class_ $class): bool
```
brambaud, line 22: Should we add a comment to explain what is considered to be a shared namespace? 🤔 

When I read the code, it seems clear to me but I don't know if it is clear for others

src/Internal/ClassHelper.php
mglaman: Eh, it works. We could name it something like `NamespaceChecker`, since it has methods for checking if the namespace "is" something.

src/Internal/ClassHelper.php
mglaman: I think the code explains it pretty well, the first two parts of the namespace must match

mglaman (approved):
Looks great, @brambaud ! I'll pause to merge to see your thoughts on s/`ClassHelper`/`NamespaceCheck`. 

I like `NamespaceCheck::isInDrupalNamespace` (or just `isDrupalNamespace`). I think it reads well

src/Internal/ClassHelper.php
brambaud: I like the `NamespaceCheck` name thanks!

brambaud:
Agreed about the naming and changes have been done! Thanks!

mglaman (approved):
This is amazing, thanks!

PR #255, opened 2021-11-27
Support dynamic return type of Drupal\Core\Url::toString

brambaud:


mglaman (approved):
You're amazing! Thanks

Issue #249, opened 2021-11-20
Make the discovered Drupal extensions accessible via a service


mglaman:
We need to make the discovered extensions available from a service, just like ServiceMap.

This would help https://github.com/mglaman/phpstan-drupal/issues/247 and also improve the existing load include return type extensions.

brambaud:
Should we enhance the existing `ExtensionDiscovery` and  `Extension` or should we create another service and/or another value object using them? 🤔

mglaman:
Fixed via https://github.com/mglaman/phpstan-drupal/pull/264

Issue #208, opened 2021-10-06
Throw an error on service names that are invoked but missing


mglaman:
Right now there are no errors if you call `entity.manager` from the service container in D9. That is because we do not error when there is a service miss. We should throw an error.

HOWEVER this would error on dynamically defined services in service providers or compiler passes.

mglaman:
Maybe we can add a list of known services to return an error on when a specific Drupal version is used.

brambaud:
To take into account the dynamically defined services, phpstan-symfony use the dumped container.
Drupal use a php array dumper because it needs the dynamic part (when installing a module for instance the container needs to be rebuilt) and do not store it in a file accessible like Symfony does, but maybe we could ask people to do an extra step by dumping the container in a file and configure phpstan-drupal if they want to use this rule? 🤔 

Just thinking out loud.

mglaman:
You're right. Usually in my experience dynamic services are usually event subscribers, or things not manually injected. So that'd be a problem

mglaman:
If this moves forward, it should only be behind a flag and opt-in. That way end users can choose to turn this on. Their code has more control and probably does not have dynamic service declarations.

PR #256, opened 2021-11-27
Support parent definition service

brambaud:
See https://symfony.com/doc/current/service_container/parent_services.html

### DIFF
--- a/src/Drupal/ServiceMap.php
+++ b/src/Drupal/ServiceMap.php
@@ -25,6 +25,10 @@ class ServiceMap
         self::$services = [];
 
         foreach ($drupalServices as $serviceId => $serviceDefinition) {
+            if (isset($serviceDefinition['parent'], $drupalServices[$serviceDefinition['parent']])) {
+                $serviceDefinition = $this->resolveParentDefinition($serviceDefinition['parent'], $serviceDefinition, $drupalServices);
+            }
+
             // @todo support factories
             if (!isset($serviceDefinition['class'])) {
                 if (isset($serviceDefinition['alias'], $drupalServices[$serviceDefinition['alias']])) {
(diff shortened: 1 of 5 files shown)


src/Drupal/ServiceMap.php
```diff
@@ -53,4 +57,30 @@ public function setDrupalServices(array $drupalServices): void
...
+            return $this->resolveParentDefinition($parentDefinition['parent'], $drupalServices[$parentDefinition['parent']], $drupalServices);
+        }
+
+        if (isset($parentDefinition['class']) && !isset($serviceDefinition['class'])) {
+            $serviceDefinition['class'] = $parentDefinition['class'];
+        }
+
+        if (isset($parentDefinition['public']) && !isset($serviceDefinition['public'])) {
+            $serviceDefinition['public'] = $parentDefinition['public'];
+        }
```
mglaman, line 82: I wish we could copy more from the actual dependency injection code. But does it just merge things? Like if we could merge the arrays

For example, this doesn't handle arguments. I guess we don't really inspect those right now.

Base example are plugin managers. Usually a custom class but parent of default plugin manager and the same arguments

mglaman (commented):
Quick initial review, curious about arguments and other things that may be inherited

src/Drupal/ServiceMap.php
brambaud: > I wish we could copy more from the actual dependency injection code. But does it just merge things? Like if we could merge the arrays

I looked at the YamlFileLoader (both on Drupal side and Symfony side) how they did it to make sure I was not making wrong assumptions before coding this.

One of the first thing they do is to either create a `ChildDefinition` if 'parent' is set or `Definition` if it is not.

Then it will define the different properties (class, shared, synthetic, lazy, public, abstract, arguments etc.. etc..) like

```php
if (isset($service['class'])) {
  $definition->setClass($service['class']);
}
```

As said in #208, I am experiencing  with dumping the container to see how it could help with everything (#247, #249 etc..)

> For example, this doesn't handle arguments. I guess we don't really inspect those right now.
>
> Base example are plugin managers. Usually a custom class but parent of default plugin manager and the same arguments

Yep, since arguments are not supported with the actual `DrupalServiceDefinition` and that we don't use them for the moment, I thought about doing another PR later to support arguments.

mglaman (approved):
Thanks! Makes sense, we don't track. Eventually we should to catch invalid arguments passed from the services.yml

Right now we can track deprecated interfaces in the constructor but not from actual argument defined (deprecated service name)

PR #260, opened 2021-11-30
Fail to map a service with a parent which overrides its parent too

brambaud:
~~The test should fail to reproduce the error~~ Now the test should pass. Sorry for the mistake 😅


mglaman (approved):
Great find on this edge case!

PR #264, opened 2021-12-01
Make the discovered Drupal extensions accessible via a service

brambaud:
See #249 

Just to open discussion

### DIFF
--- a/src/Drupal/ExtensionMap.php
+++ b/src/Drupal/ExtensionMap.php
@@ -4,13 +4,13 @@ namespace mglaman\PHPStanDrupal\Drupal;
 
 final class ExtensionMap
 {
-    /** @var Extension[]  */
+    /** @var array<string, Extension>  */
     private static $modules = [];
 
-    /** @var Extension[]  */
+    /** @var array<string, Extension>  */
     private static $themes = [];
 
-    /** @var Extension[]  */
+    /** @var array<string, Extension>  */
     private static $profiles = [];
 
     /**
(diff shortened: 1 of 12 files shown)


src/Drupal/ExtensionMap.php
```diff
@@ -0,0 +1,54 @@
...
+                }
+
+                return $name;
+            }, $info['dependencies'] ?? []);
+
+            self::$modules[$module->getName()] = new ModuleExtension(
+                $module->getName(),
+                \sprintf('%s/%s', $drupalRoot, $module->getPathname()),
+                $dependencies
+            );
```
mglaman, line 51: Drupal core does this... somewhere. But I think it sets dynamic properties on the extension object. So that's a bummer. Maybe we could extend our fork of `Extension` to allow lazy processing of the YAML data so we don't need to decode everything up front?

src/Drupal/ModuleExtension.php
```diff
@@ -0,0 +1,55 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Drupal;
+
+final class ModuleExtension
```
mglaman, line 5: I think it'd be good to just add these details to the existing `Extension` object. Originally the idea was to keep a more "pure" fork of the Drupal code... but I have given up on that :)

src/Drupal/ExtensionMap.php
mglaman: It's set here: https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/lib/Drupal/Core/Extension/ExtensionList.php#L316

```php
  protected function createExtensionInfo(Extension $extension) {
    $info = $this->infoParser->parse($extension->getPathname());
```

So we have extension discovery but not the list forked. I think it's fine just encapsulating that info parsing into the Extension class and lazy-loading it as needed.

src/Drupal/ModuleExtension.php
brambaud: I think we should try to have different objects because modules, themes and profiles have different domain behaviours.

They have different configurations and informations inside there `.info.yml`.

For instance themes have specific parameters like `regions`, `libraries`, `engine` whereas profiles have `install`, `distribution` and modules have `test_dependencies` etc..

It will be simple to work and avoid a god object. WDYT?

src/Drupal/ExtensionMap.php
brambaud: The issue with the extension list and the info parser is that they do not parse all the information. When I looked at it the info parser only handle `core_version_requirement`, `package`, `core_incompatible`, and `version`.

Each time we have to look at the specific extension list for each project type like `https://git.drupalcode.org/project/drupal/-/blob/9.3.x/core/lib/Drupal/Core/Extension/ModuleExtensionList.php`.

I will try to look at this tomorrow or this weekend to have something less POC 😛

mglaman: I don't think this library needs to inherit whatever decisions core made. We can try those properties in our Extension object since we know _all_ of them have it.

src/Drupal/ModuleExtension.php
mglaman: Will we be inspecting the internal properties, though? I guess down the line. But right now the main thing is for properties that are the same across any extension type.

src/Drupal/ExtensionMap.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @param Extension[] $modules
+     * @param Extension[] $themes
+     * @param Extension[] $profiles
+     */
+    public function setExtensions(array $modules, array $themes, array $profiles): void
+    {
+        self::$modules = $modules;
+        self::$themes = $themes;
+        self::$profiles = $profiles;
```
mglaman, line 64: The incoming data is _not_ keyed by extension name, so all of these getters will return null.

mglaman (changes requested):
I expanded this to reduce code in the load include rules. I noticed that all of the ExtensionMap getters are null since data is not keyed by module name when set. Pushing fix with refactored load includes code

mglaman (approved):

PR #265, opened 2021-12-01
Allow ::referencedEntities on field items

eiriksm:
OK, so this PR adds a bit to the legacy that is allowed things on a field item.

Currently we just assume that if an object is of the type FieldItemListInterface (or super type of) then we can access the properties entity, value and target_id.

Technically this might be true, as accessing ->entity on a text item would probably just give you null? But my claim is that if we are blindly allowing things that might be from another field, we might as well allow methods that might be on another field. For example ::referencedEntities.

Ideally I would see some sort of elaborate entity field mapping thing, where you can actually specify that field_coupon would be an entityreference item (and its class) but that's more of a refactor than this PR (and my time) allows. So here is a stab at allowing that specific method as well, together with those specific properties that we already allow.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -282,6 +282,9 @@ services:
 		class: mglaman\PHPStanDrupal\Reflection\EntityFieldsViaMagicReflectionExtension
 		tags: [phpstan.broker.propertiesClassReflectionExtension]
 	-
+		class: mglaman\PHPStanDrupal\Reflection\EntityFieldMethodsViaMagicReflectionExtension
+		tags: [phpstan.broker.methodsClassReflectionExtension]
+	-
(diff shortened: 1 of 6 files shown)


mglaman:
@eiriksm sorry I didn't get to this yet, I will tackle it soon to help it land!

eiriksm:
Ah no worries, I was not expecting you to review or in any other way tackle this when the tests are not passing. I just have not found the time to finish it

mglaman:
@eiriksm squashed this up. Thanks for the add!

eiriksm:
Wow thanks! ❤️

mglaman:
No problem! Hopefully one day I can sort out the mess of entity field properties and field item lists

PR #279, opened 2021-12-22
Add test for setting an empty field's first item to null

### LINKED ISSUE #228: Add test for setting an empty field's first item to null
Spinning from https://git.drupalcode.org/project/drupal/-/merge_requests/454#note_55009

Drupal.org issue: https://www.drupal.org/project/drupal/issues/3254966

```patch
-    $entity->field_test[0] = NULL;
+     $entity->field_test = NULL;
```

Without this fix, PHPStan crashes when running against Drupal core

```
Internal error: Internal error: Cannot assign an empty string to a string  
offset in file                                                             
/var/www/html/core/modules/link/tests/src/Kernel/LinkItemTest.php          
Run PHPStan with --debug option and post the stack trace to:               
https://github.com/phpstan/phpstan/issues/new?template=Bug_report.md       
Child process error (exit code 1):
```

mglaman:
This verifies a Drupal core bug when processing LinkItemTest

Fixes #228﻿

### DIFF
--- a/tests/src/LinkItemTestBug3254966Test.php
+++ b/tests/src/LinkItemTestBug3254966Test.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Tests;
+
+use mglaman\PHPStanDrupal\Tests\Rules\DummyRule;
+
+class LinkItemTestBug3254966Test extends DrupalRuleTestCase
+{
+    protected function getRule(): \PHPStan\Rules\Rule
+    {
+        return new DummyRule();
+    }
+
+    public function testCrash(): void
+    {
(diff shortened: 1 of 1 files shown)


mglaman:
Screen Shot 2021-12-22 at 3 28 32 PM

This is why it's erroring. PHPStan has held onto the value when it was set to a string.

PHPStan is not aware that the entity property is a FieldItemList which is \ArrayAccess but with magical getters that will return the first value.

mglaman:
The "bug": NodeScopeResolver::processAssignVar thinks the that field value is a string which no longer is a super of \ArrayAccess and it forgets that it is a FieldItemList

mglaman:
Parent issue was closed without this being merged, as it was apparently fixed already.

PR #282, opened 2021-12-28
Add phpstan prophecy suggestion

### LINKED ISSUE #263: Add jangregor/phpstan-prophecy as a suggested package

ptt-homme:
Hi @mglaman

This PR fixes #263. It also removes `phpstan-deprecation-rule` from the list, as the package is part of the dev dependencies.

Hope this helps!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,10 @@
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
-    "suggest": {},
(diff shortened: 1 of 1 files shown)


mglaman:
Dev dependencies only get installed when it's the root package. We need to keep it as a suggestion for sites that I stalk this package because it's still not a direct dependency

PR #292, opened 2022-01-06
Fix/render callback

alexpott:
Objects that implement Drupal\Core\Render\Element\RenderCallbackInterface are allowed to use their methods as render callbacks.


mglaman (approved):
Thanks!!

Issue #299, opened 2022-01-11
Calling getStorage with a constant does not resolve entity mapped storage class


dpi:
Calling getStorage where the first parameter (`$entity_type_id`) is a constant does not resolve the correct storage class.

Given.

```yaml
parameters:
	drupal:
		entityMapping:
			my_custom_config_entity_type:
				class: Drupal\my_custom_config_entity_type\Entity\MyCustomConfigEntityType
				storage: Drupal\Core\Config\Entity\ConfigEntityStorage
```

PHPStan does not resolve the class for getStorage properly when this is called.

```php
$entities = \Drupal::entityTypeManager()->getStorage(MyCustomConfigEntityType::ENTITY_TYPE_ID)->loadMultiple();
```

`$entities` is `\Drupal\Core\Entity\EntityInterface[]`

Expected

`$entities` is `\Drupal\my_custom_config_entity_type\Entity\MyCustomConfigEntityType[]`

mglaman:
@dpi forgot to say, I'll have this tagged tomorrow after my live stream dev session

dpi:
Good to hear, thanks!

PR #300, opened 2022-01-11
Calling getStorage with a constant does not resolve entity mapped storage class

dpi:
Fixes mglaman/phpstan-drupal#299


src/Type/EntityTypeManagerGetStorageDynamicReturnTypeExtension.php
```diff
@@ -74,13 +75,15 @@ public function getTypeFromMethodCall(
         if ($arg1 instanceof Concat) {
             return $returnType;
         }
-        if (!$arg1 instanceof String_) {
+
+        $type = $scope->getType($arg1);
+        if ($type instanceof ConstantStringType) {
+            $entityTypeId = $type->getValue();
```
mglaman, line 81: Ah! Great catch, we didn't use the scope to get the type.

mglaman (approved):
Thank you for this catch!

dpi:
Thanks @mglaman

PR #306, opened 2022-01-17
Relax constraints

alexpott:


alexpott:
This will allow this to be installed on Drupal 10 not that Drupal 10 has been updated to use SF 6. I've relaxed all the constraints.

mglaman (approved):
:shipit:

PR #319, opened 2022-01-31
BrowserTestBaseDefaultThemeRule namespace check

### LINKED ISSUE #318: Improve performance of BrowserTestBaseDefaultThemeRule
Spinning out from https://github.com/mglaman/phpstan-drupal/issues/316

When scanning `core/modules/action` which has 21 files analyzed, BrowserTestBaseDefaultThemeRule is invoked 19 times. Scanning the entire files only took 48.4s but that rule accounted for 23.8s seconds! That is 49.24% of the time.

Screen Shot 2022-01-31 at 2 28 55 PM

I assume it has to be this first check to see if the class extends BrowserTestBase

```
       $classType = $scope->resolveTypeByName($node->namespacedName);
        assert($classType instanceof ObjectType);

        $browserTestBaseType = new ObjectType('Drupal\\Tests\\BrowserTestBase');
        if (!$browserTestBaseType->isSuperTypeOf($classType)->yes()) {
            return [];
        }
```

There needs to be a lighter preflight check besides digging into reflection.

mglaman:
Uses a namespace check before performing reflections in BrowserTestBaseDefaultThemeRule. This seems to shave 3-5s off of the sample set of analyzing the `action` module.

Fixes #318

### DIFF
--- a/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
+++ b/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
@@ -29,10 +29,6 @@ final class BrowserTestBaseDefaultThemeRule implements Rule
         if ($node->namespacedName === null) {
             return [];
         }
-        $testNamespacePart = (string) $node->namespacedName->slice(1, 3);
-        if ($testNamespacePart !== 'Tests\\Functional' && $testNamespacePart !== 'Tests\\FunctionalJavascript') {
-            return [];
-        }
 
         // Only inspect tests.
         // @todo replace this str_ends_with() when php 8 is required.
--- a/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
+++ b/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
@@ -29,6 +29,10 @@ final class BrowserTestBaseDefaultThemeRule implements Rule
         if ($node->namespacedName === null) {
             return [];
         }
+        $testNamespacePart = (string) $node->namespacedName->slice(1, 3);
+        if ($testNamespacePart !== 'Tests\\Functional' && $testNamespacePart !== 'Tests\\FunctionalJavascript') {
+            return [];
+        }
 
         $classType = $scope->resolveTypeByName($node->namespacedName);
         assert($classType instanceof ObjectType);


src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
```diff
@@ -29,6 +29,10 @@ public function processNode(Node $node, Scope $scope): array
         if ($node->namespacedName === null) {
             return [];
         }
+        $testNamespacePart = (string) $node->namespacedName->slice(1, 3);
+        if ($testNamespacePart !== 'Tests\\Functional' && $testNamespacePart !== 'Tests\\FunctionalJavascript') {
+            return [];
+        }
```
mglaman, line 35: 🤦 it's `Drupal\Tests\{module\`. So the cost savings is there. Just wrong code

mglaman:
The work in https://github.com/mglaman/phpstan-drupal/pull/323 is way better

mglaman:
@neclimdul okay, I get what went wrong and why GHA didn't run.

mglaman:
lol FML I can't do integration tests because Drupal core requires phpstan-drupal

```
Problem 1
    - Root composer.json requires mglaman/phpstan-drupal *@dev, found mglaman/phpstan-drupal[dev-f8ce5aaada2[21](https://github.com/mglaman/phpstan-drupal/runs/5170920882?check_suite_focus=true#step:5:21)423942ae6f3f7777c25c515df07] but these were not loaded, likely because it conflicts with another require.
  Problem 2
    - drupal/core-dev is locked to version 10.0.x-dev and an update of this package was not requested.
    - drupal/core-dev 10.0.x-dev requires mglaman/phpstan-drupal ^1.1.9 -> satisfiable by mglaman/phpstan-drupal[1.1.9, 1.1.10] from composer repo (https://repo.packagist.org/) but mglaman/phpstan-drupal[dev-f8ce5aaada221423942ae6f3f7777c25c515df07] from path repo (/home/runner/work/phpstan-drupal/phpstan-drupal) has higher repository priority. The packages from the higher priority repository do not match your constraint and are therefore not installable. That repository is canonical so the lower priority repo's packages are not installable. See https://getcomposer.org/repoprio for details and assistance.
``

mglaman:
Opening a new PR to see if that can be dropped.

mglaman:
10.0.x is allowed to fail, and there's actually a lot to sort out. So I'm just merging this1

PR #323, opened 2022-02-12
Optimize BrowserTestBaseDefaultThemeRule

neclimdul:
Avoid inspecting classes we know aren't tests or Functional/Browser tests.

Fixes https://github.com/mglaman/phpstan-drupal/issues/318

### DIFF
--- a/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
+++ b/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
@@ -30,6 +30,28 @@ final class BrowserTestBaseDefaultThemeRule implements Rule
             return [];
         }
 
+        // Only inspect tests.
+        // @todo replace this str_ends_with() when php 8 is required.
+        if (0 !== substr_compare($node->namespacedName->getLast(), 'Test', -4)) {
+            return [];
+        }
+
+        // Do some cheap preflight tests to make sure the class is in a
+        // namespace that makes sense to inspect.
+        $parts = $node->namespacedName->parts;
+        // The namespace is too short to be a test so skip inspection.
+        if (count($parts) < 3) {
+            return [];
+        }
+        // If the 4th component matches it's a module test. If the 2nd, core.
+        if ($parts[3] !== 'Functional'
+            && $parts [3] !== 'FunctionalJavascript'
+            && $parts[1] !== 'FunctionalTests'
+            && $parts[1] !== 'FunctionalJavascriptTests') {
+            return [];
+        }
+
+
         $classType = $scope->resolveTypeByName($node->namespacedName);
         assert($classType instanceof ObjectType);
 


neclimdul:
@mglaman maybe something like this?

Bonus, it fixes this test asserting the defaultTheme on base tests which fixes two entries in the core baseline.

src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
```diff
@@ -30,6 +30,26 @@ public function processNode(Node $node, Scope $scope): array
             return [];
         }
 
+        // Only inspect tests.
+        if (!str_ends_with($node->namespacedName->getLast(), 'Test')) {
+            return [];
+        }
```
mglaman, line 36: This is a PHP 8.0 function, so we'd need the old school `strpos` or something.

mglaman (changes requested):
I don't know why it isn't showing me the prompt to run GitHub Actions.

If we did, I bet we'd see the crash on PHP 7 that CircleCI is running. This is better than what I cooked up in https://github.com/mglaman/phpstan-drupal/pull/319 and forgot to follow through with.

src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
neclimdul: its in `symfony/polyfill-php80` so it should be fine.

mglaman: nice! I never consider this

neclimdul: well... i say that. its a indirect dependency so that's _technically_ not a good solution. replace it.

src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
```diff
@@ -30,6 +30,27 @@ public function processNode(Node $node, Scope $scope): array
...
+        // The namespace is too short to be a test so skip inspection.
+        if (count($parts) < 3) {
+            return [];
+        }
+        // If the 4th component matches it's a module test. If the 2nd, core.
+        if ($parts[3] !== 'Functional' &&
+            $parts [3] !== 'FunctionalJavascript' &&
+            $parts[1] !== 'FunctionalTests' &&
+            $parts[1] !== 'FunctionalJavascriptTests') {
+            return [];
```
mglaman, line 51: These need to be split up, I think. Because it's `&&`. So should it actually be all `||`? So that the first `true` exits the conditional?

Also, I've taken the style from PHPStan of putting `&&` or `||` first, so:

```php
        if ($parts[3] !== 'Functional'
            || $parts [3] !== 'FunctionalJavascript'
            || $parts[1] !== 'FunctionalTests'
            || $parts[1] !== 'FunctionalJavascriptTests') {
```

neclimdul: I think the logic is actually correct but it is just one of those confusing inversions because we're providing a firewall.

We want to continue if one of the namespaces match which is to say `a == 'Functional' || a == 'FunctionalJavascript' || b == 'FunctionalTests' || b == 'FunctionalJavascriptTests'`

and the logic inversion of `a ==  b || c == d` is `a != b && c != d` so we get the && rule.

I'll change the formatting though.

neclimdul:
ugh, failure was me failing to convert ends_with correctly.

mglaman (approved):
well, I guess GHA is a lost cause, I'll merge it up anyways!

mglaman:
Thank you @neclimdul! This is a great if it helps fix a false-positive in Drupal core's baseline, and it provides a bump in performance.

PR #327, opened 2022-02-13
Provide a stub for twig_escape_filter

### LINKED ISSUE #310: Function twig_escape_filter not found
The function lives in \Twig\Extension\EscaperExtension. Not sure why it is not discovered.

We just need to make sure that file gets autoloaded in scanning. It does when Drupal runs for some reason

neclimdul:
twig_escape_filter is a hidden function inside a class file so not discovered
though normal discovery or autoloading so providing scan stub allows its use
to be analyzed.

Fixes #310

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -11,6 +11,8 @@ parameters:
 		- install
 		- profile
 		- engine
+	scanFiles:
+		- stubs/Twig/functions.stub
 	drupal:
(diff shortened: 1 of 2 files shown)


neclimdul:
I couldn't recreate it outside the Drupal repository either but this fixes it.

mglaman (approved):
nice. thank you!

PR #345, opened 2022-02-18
drupal/core:10.0.x requires PHP 8.1

mglaman:


mglaman:
PHPUnit will fail until https://github.com/drush-ops/drush/issues/5077. Drush isn't used as part of the integration builds.

PR #351, opened 2022-02-23
Rule for ConditionManager::createInstance() when 'context' is passed as array key

### LINKED ISSUE #337: Rule for ConditionManager::createInstance() when  'context' is passed as array key
CR here: https://www.drupal.org/node/3120980

mglaman:
Fixes #337


mglaman:
Blah, I forgot to add a minor version check with this. We'll see if that causes problems.

PR #355, opened 2022-02-25
Entity storage with type hint prevent inferring entity query execute return type

brambaud:
For the moment just commit tests that should fail to reproduce the issue


brambaud:
`EntityTypeManagerGetStorageDynamicReturnTypeExtension` seems to return the correct type, i.e. `ContentEntityStorageType`, even with the type hint but in `EntityStorageDynamicReturnTypeExtension::getTypeFromMethodCall()` the `$callerType` is just an `ObjectType` instead of an instance of `EntityStorageType`.
Not sure why yet 🤔

mglaman:
That's interesting. I'll dig into this on my Wednesday live stream unless you figure it out beforehand

brambaud:
I can't figure out why we have this behaviour 😞 
@mglaman If you find out please let me now!

I pushed a quick workaround to test the idea for the query only that seems to work but we need more test cases.

src/Type/EntityStorage/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -60,6 +61,18 @@ public function getTypeFromMethodCall(
...
+            if (!$callerType instanceof ObjectType) {
+                return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+            }
+
+            // Workaround because cannot figure out why the caller type is not an EntityStorageType
+            // when it has been type hinted.
+            // Instead, we try to infer the type, i.e. ContentEntityStorageType or ConfigEntityStorageType, here.
+            // @todo: we should definitively look for other cases that getQuery.
+            if ($methodReflection->getName() === 'getQuery') {
+                return $this->getReturnTypeForGetQueryMethod($methodReflection, EntityStorageHelper::getTypeFromStorageObjectType($callerType));
```
brambaud, line 73: If we really don't find another way to fix the issue, instead of this we should refactor the code to do something like:

```php
$callerType = $scope->getType($methodCall->var);

if (!$callerType instanceof EntityStorageType && !$callerType instanceof ObjectType) {
    return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
}

$entityStorageType = $callerType instanceof EntityStorageType ? $callerType : EntityStorageHelper::getTypeFromStorageObjectType($callerType);

// refactor the rest to use $entityStorageType instead of $callerType
```

src/Type/EntityStorage/EntityStorageHelper.php
```diff
@@ -0,0 +1,39 @@
...
+
+        $storageType = new ObjectType($storageClassName);
+        if ((new ObjectType(EntityStorageInterface::class))->isSuperTypeOf($storageType)->no()) {
+            return null;
+        }
+        if ((new ObjectType(ConfigEntityStorageInterface::class))->isSuperTypeOf($storageType)->yes()) {
+            return new ConfigEntityStorageType('', $storageClassName);
+        }
+        if ((new ObjectType(ContentEntityStorageInterface::class))->isSuperTypeOf($storageType)->yes()) {
+            return new ContentEntityStorageType('', $storageClassName);
```
brambaud, line 35: If we continue this way we should find a way to pass the real entity type id

mglaman: I tried some code which had the entity type ID in the class and it ended up becoming cached.

mglaman:
Digging in and found \PHPStan\PhpDoc\TypeNodeResolverExtension::EXTENSION_TAG (`phpstan.phpDoc.typeNodeResolverExtension`) extension. Allow enriching phpDoc `@var` tags. However I don't think it has enough context for enrichment like we do.

mglaman:
The `phpDoc.typeNodeResolverExtension` may not be a great idea as it runs on `@return` statements as well.

mglaman:
Tracked it down to \PHPStan\Analyser\NodeScopeResolver::processExprNode

```
            $vars = $this->getAssignedVariables($expr->var);
            if (\count($vars) > 0) {
                $varChangedScope = \false;
                $scope = $this->processVarAnnotation($scope, $vars, $expr, $varChangedScope);
```

The assigned variables are correct, but they get "broken" in `processVarAnnotation`
Screen Shot 2022-03-09 at 3 28 37 PM

mglaman:
So I spent a bunch of time on this and got no where closer. I made a phpDoc extension and that didn't help for `instanceof`. But copying here.

```
<?php declare(strict_types=1);

namespace mglaman\PHPStanDrupal\Type\EntityStorage;

use Drupal\Core\Config\Entity\ConfigEntityStorageInterface;
use Drupal\Core\Entity\ContentEntityStorageInterface;
use Drupal\Core\Entity\EntityStorageInterface;
use PHPStan\Analyser\NameScope;
use PHPStan\PhpDoc\TypeNodeResolverExtension;
use PHPStan\PhpDocParser\Ast\Type\IdentifierTypeNode;
use PHPStan\PhpDocParser\Ast\Type\TypeNode;
use PHPStan\Type\Type;

final class EntityStorageTypeNodeResolverExtension implements TypeNodeResolverExtension
{

    public function resolve(TypeNode $typeNode, NameScope $nameScope): ?Type
    {
        if (!$typeNode instanceof IdentifierTypeNode) {
            return null;
        }
        $stringName = $nameScope->resolveStringName($typeNode->name);
        if (!is_subclass_of($stringName, EntityStorageInterface::class)) {
            return null;
        }

        if (is_subclass_of($stringName, ConfigEntityStorageInterface::class)) {
            return new ConfigEntityStorageType('', $stringName);
        }
        if (is_subclass_of($stringName, ContentEntityStorageInterface::class)) {
            return new ContentEntityStorageType('', $stringName);
        }
        return new EntityStorageType('', $stringName);
    }
}
```

The problem is that this breaks ALL dynamic return type method extensions as we have them. 💩

src/Type/EntityStorage/EntityStorageHelper.php
mglaman: EDIT: Nevermind that comment, I was thinking of something else. We do this elsewhere already.

mglaman:
@brambaud maybe we just need to provide stubs for the storage types with the correct return types. That seems a heck of a lot easier when someone has used an `instanceof` or `@var`

mglaman:
@brambaud stubs didn't work. I don't know _what_ was happening but stubbing `create` on `NodeStorage` or `NodeStorageInterface` did nothing. Also, I realized that stubbing entity query objects would be impossible.

I've updated the code to allow resolving an entity type based on the storage class. So if the storage variable is type hinted correctly and the entity mapping is setup, we can resolve the correct entity data.

src/Type/EntityStorage/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -58,39 +61,24 @@ public function getTypeFromMethodCall(
...
-            return new ArrayType(new IntegerType(), $type);
-        }
         if ($methodReflection->getName() === 'getQuery') {
             $returnType = ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
             if (!$returnType instanceof ObjectType) {
                 return $returnType;
             }
-            if ($callerType instanceof ContentEntityStorageType) {
+
+            if ((new ObjectType(ContentEntityStorageInterface::class))->isSuperTypeOf($callerType)->yes()) {
```
mglaman, line 74: This is impossible to grok from the diff. All I did was move `getQuery` above the check for the entity type class. Because we don't really care about the entity type class if we have an object type that's a storage class. Then we can determine the kind of query will be executed, which does not require knowing the entity type class.

mglaman: I'm thinking we should do this in its own PR. So make `EntityStorageGetQueryDynamicReturnTypeExtension` to just handle this `getQuery` method since we do not require the entity type class. Then the `EntityStorageDynamicReturnTypeExtension` only handles methods which return a single instance or array of the entity type class.

src/Type/EntityStorage/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -100,6 +88,35 @@ public function getTypeFromMethodCall(
...
-        return $type;
+        if (!$callerType instanceof EntityStorageType) {
+            $resolvedEntityType = $this->entityDataRepository->resolveFromStorage($callerType);
+            if ($resolvedEntityType === null) {
+                return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
+            }
+            $type = $resolvedEntityType->getClassType();
+        } else {
+            $type = $this->entityDataRepository->get($callerType->getEntityTypeId())->getClassType();
+        }
```
mglaman, line 99: This is the new trick for resolving the entity type. This should stay in this PR, I believe as it is specifically used for resolving from `/** @var */`.

src/Type/EntityStorage/EntityStorageDynamicReturnTypeExtension.php
```diff
@@ -100,6 +88,35 @@ public function getTypeFromMethodCall(
...
+            }
+
+            return new ArrayType(new IntegerType(), $type);
+        }
+
+        if ($methodReflection->getName() === 'create') {
+            return $type;
+        }
+
+        return ParametersAcceptorSelector::selectSingle($methodReflection->getVariants())->getReturnType();
```
mglaman, line 120: I added this to be more explicit. I know we restrict what method names are handled, but it felt weird just dangling the entity class as the return type.

mglaman (approved):

PR #356, opened 2022-03-09
The entity property can return NULL

larowlan:
If you delete the entity that is referenced by an ER field, the entity property will return NULL

Code needs to be aware of that and handle it.

Without this change, code that is being defensive for when entity is NULL shows linting errors on phpstan level 6


mglaman (approved):
Ah, good catch. Thanks!

PR #382, opened 2022-04-07
Entity query count and accessCheck order should not matter

### LINKED ISSUE #381: Missing access check reports false negatives
The check added in #378 is quite particular about how/where the `accessCheck` function is called and can report errors on queries that are actually calling it.

In particular, any `count()` queries error unless the `accessCheck` call is _after_ the `count()` call.

This produces an error:
```
$query = \Drupal::entityTypeManager()->getStorage('node')
      ->getQuery()
      ->accessCheck(FALSE)
      ->count();
$count = (int) $query->execute();
```
and this:
```
$query = \Drupal::entityTypeManager()->getStorage('node')
      ->getQuery()
      ->count();
$query->accessCheck(FALSE);
$count = (int) $query->execute();
```

This doesn't:
```
$query = \Drupal::entityTypeManager()->getStorage('node')
      ->getQuery()
      ->count()
      ->accessCheck(FALSE);
$count = (int) $query->execute();
```

There's also many other examples, most seem to be around calling `$query->accessCheck()` on its own line.

brambaud:
Fixes #381

### DIFF
--- a/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityQueryWithAccessRule.php
+++ b/tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityQueryWithAccessRule.php
@@ -22,4 +22,28 @@ final class EntityQueryWithAccessRule
             ->condition('field_test', 'foo', '=')
             ->execute();
     }
+
+    public function count(): void
+    {
+        \Drupal::entityTypeManager()->getStorage('node')
+            ->getQuery()
+            ->accessCheck(false)
+            ->count()
+            ->execute();
+    }
+
+    public function bug381CountOrderShouldNotMatter(): void
+    {
+        $query = \Drupal::entityTypeManager()->getStorage('node')
+            ->getQuery()
+            ->accessCheck(FALSE)
+            ->count();
+        $count = (int) $query->execute();
+
+        $query = \Drupal::entityTypeManager()->getStorage('node')
+            ->getQuery()
+            ->count()
+            ->accessCheck(FALSE);
+        $count = (int) $query->execute();
+    }
 }
--- a/src/Type/EntityQuery/EntityQueryCountType.php
+++ b/src/Type/EntityQuery/EntityQueryCountType.php
@@ -2,8 +2,6 @@
 
 namespace mglaman\PHPStanDrupal\Type\EntityQuery;
 
-use PHPStan\Type\ObjectType;
-
 /**
  * Type used to represent an entity query instance as count query.
  */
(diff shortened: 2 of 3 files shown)


tests/fixtures/drupal/modules/phpstan_fixtures/src/EntityQueryWithAccessRule.php
```diff
@@ -22,4 +22,34 @@ public function bar(): void
...
+        $query = \Drupal::entityTypeManager()->getStorage('node')
+            ->getQuery()
+            ->accessCheck(FALSE)
+            ->count();
+        $count = (int) $query->execute();
+
+        $query = \Drupal::entityTypeManager()->getStorage('node')
+            ->getQuery()
+            ->count();
+        $query->accessCheck(FALSE);
```
brambaud, line 46: With `$query->accessCheck(FALSE);` it's ok.

The type of `$query` is not changed after calling `accessCheck` if we don't assign it.
I think we may have the issue elsewhere 🤔

brambaud: Does a type-specifying extension could help? 🤔

mglaman: 😬 I'm not sure. This is getting _very_ tricky. I wonder if any other extensions support this kind of wild magic. I reviewed Laravel's and there is some magic in models, but nothing like this

brambaud: Type-specifying won't help here, there are for assert call like

It seems the whole idea with specific types like `EntityQueryType` and co does not help for those cases 🤔 
I checked in phpstan-doctrine to know how they handle the `rowCount` method and other like that and for the moment they only support chaining methods (https://github.com/phpstan/phpstan-doctrine/blob/1.3.x/src/Type/Doctrine/DBAL/QueryBuilder/QueryBuilderExecuteMethodExtension.php)

brambaud: What do you think about doing this in another PR and merge the fix for the other case? I could quickly update the PR to do this

mglaman: I'm all for quick fixes for this

mglaman: @brambaud interesting, for what phpstan-doctrine does, I wonder if we should try that out in instance misses from `($methodName === 'execute')`

PR #391, opened 2022-04-13
Add Drush command for configuring PHPStan

mglaman:


src/Drush/Commands/PhpstanDrupalDrushCommands.php
```diff
@@ -0,0 +1,58 @@
...
+    public function setup($options = ['file' => null]): void
+    {
+        $parameters = [
+            'level' => 2,
+            'paths' => [
+                // @todo use drupal-finder for docroot
+                'web/modules/custom',
+                'web/themes/custom',
+                'web/profiles/custom',
+            ],
```
mglaman, line 24: Use drupal-finder instead of hardcoding `web`

src/Drush/Commands/PhpstanDrupalDrushCommands.php
```diff
@@ -0,0 +1,58 @@
...
+            // @todo can we have this override _everything_ phpstan-drupal provides? or is it a merge.
+            'entityMapping' => [
+            ],
```
mglaman, line 27: Look into this, should we drop default entity type mappings, otherwise?

src/Drush/Commands/PhpstanDrupalDrushCommands.php
```diff
@@ -0,0 +1,58 @@
...
+parameters:
+
+NEON;
+;
+        foreach ($parameters as $key => $value) {
+            $output .= "$key:";
+            if (is_array($value)) {
+                $output .= "\n\t";
+            }
+        }
```
mglaman, line 49: Evaluate what to do here: we need to dump as NEON but that would add a dependency, which could be frowned upon by the Drupal core maintainers.

mglaman:
Instead of being used to dump `phpstan.neon`, could be a supplemental file that gets included into the phpstan.neon for `drupal` related items.

ondrejmirtes:
You don’t need to use nette/neon to write the file, you can output it like a human would, by echoing tabs and newlines 😊

mglaman:
> You don’t need to use nette/neon to write the file, you can output it like a human would, by echoing tabs and newlines 😊

😅 Yeah, thinking back it's not that bad. Especially since it's just "is string or array?" But apparently it was too much for my end of day brain

mglaman:
Fails due to Symfony v4 and v6 compat, see https://github.com/drush-ops/drush/pull/5108/files#diff-fcb54a74a331c6112bd5cf5e32f08a4d0a39f42c16833b899b4fac0339d453c7

Need to call \Drush\Preflight\Preflight::loadSymfonyCompatabilityAutoloader – Drush 11 only!

batkor:
Hi.
Please see this my commit https://github.com/batkor/phpstan-drupal/commit/b6afccec70773f12cb230be5f73cb7569e24764b
I stard create drushGenerator command for generate drupal meta data.
And use metadata map on rules.

i think need generate another metadata and integrate to on remain rules.

PR #410, opened 2022-05-24
Fix PHPStan 1.7.0 incompatibilities

mglaman:


mxr576:
phpstan/phpstan version constraint should be also updated in composer.json,should not be? 

Also, if similar changes are expected in upstream, maybe it is a good idea being specific about minor releases that this plugin should work otherwise Composer is going install incomplete versions, my suggestion is ~1.7.0 instead of ^1.7.

PR #419, opened 2022-05-28
Replace AccessibleReturnTypeExtension with stub for AccessibleInterface

### LINKED ISSUE #414: Entity ::access method fails to detect object will be returned
Spreads

```php
$accessArgs = ['delete', NULL, TRUE];
$access = $entity->access(...$accessArgs);
$access->isForbidden();
```

Named args

```php
$access = $entity->access('delete', return_as_object: TRUE);
$access->isForbidden();
```

Results in:

> phpstan: Cannot call method isForbidden() on bool.

Seems like whatever checking third arg == TRUE will return object needs a little more intelligence about whats being passed to `access`.

mglaman:
Fixes #414


tests/src/Type/data/bug-414.php
```diff
@@ -0,0 +1,33 @@
...
+    $access = $entity->access(...$accessArgs);
+    assertType('bool', $access);
+
+    $accessArgs = ['delete', null, false];
+    $access = $entity->access(...$accessArgs);
+    assertType('bool', $access);
+
+    $accessArgs = ['delete', null, true];
+    $access = $entity->access(...$accessArgs);
+    assertType(AccessResultInterface::class, $access);
```
mglaman, line 25: @dpi so, this still fails even with the new stub. I haven't dug in via xdebug, yet. It's like the parameters are not analyzed correctly when using spreads.

mglaman: Replicated on phpstan.dev: https://phpstan.org/r/7afd799d-3169-4ac9-a69b-7debbdc9d3d5

Issue #397, opened 2022-04-30
Provide `service-id-string` for service ID parameters


mglaman:
We can provide custom PHPStan PHPDoc types. This could help us replace the dynamic return type extension DrupalServiceDynamicReturnTypeExtension. This can allow us to validate known service names as arguments.

```
@phpstan-param service-name-string $id
```

Then, maybe, we can tie it into templates for returning the class object type. 

Such as the following in a stub of `\Drupal::service`

```
@phpstan-param service-name-string<T> $id
@return T
```

If our extension to provide `service-name-string` is able to provide the value of `T` for the template, it might just work.

mxr576:
Okay, so this is not yet implemented :) 
(I have just had a need for something similar.)

mglaman:
I started working on it and dead-ended. I may try to pick it back up. It's been a while

PR #434, opened 2022-06-14
add support for class string service ids

laytan:
According to [this docs page](https://www.drupal.org/docs/drupal-apis/services-and-dependency-injection/services-and-dependency-injection-in-drupal-8#s-defining-services-by-fully-qualified-name-of-php-namespace) using the fully qualified class name when interacting with services is a supported approach.

This PR adds the correct return type hint when this approach is used.

### DIFF
--- a/src/Type/ContainerDynamicReturnTypeExtension.php
+++ b/src/Type/ContainerDynamicReturnTypeExtension.php
@@ -4,7 +4,10 @@ namespace mglaman\PHPStanDrupal\Type;
 
 use mglaman\PHPStanDrupal\Drupal\DrupalServiceDefinition;
 use mglaman\PHPStanDrupal\Drupal\ServiceMap;
+use PhpParser\Node;
+use PhpParser\Node\Expr\ClassConstFetch;
 use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Scalar\String_;
 use PhpParser\Node\VariadicPlaceholder;
 use PHPStan\Analyser\Scope;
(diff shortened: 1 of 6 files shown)


src/Type/ContainerDynamicReturnTypeExtension.php
```diff
@@ -73,4 +75,18 @@ public function getTypeFromMethodCall(
...
+    protected function getServiceId(Node $arg1): ?string
+    {
+        if ($arg1 instanceof String_) {
+            // @todo determine what these types are.
+            return $arg1->value;
+        }
+
+        if ($arg1 instanceof ClassConstFetch && $arg1->class instanceof FullyQualified) {
+            return (string) $arg1->class;
+        }
```
mglaman, line 88: I wonder if there's anyway to leverage PHPStan type system, which may abstract away some of this logic. But it may not be as accessible in these extensions

laytan: I saw this issue #397 where you are adding a 'service-id-string' type which, correct me if I'm wrong, looks to make this plugin obsolete?

If so, I think this is a good starting point to get support for the described feature into phpstan-drupal, including tests that make following implementations think of the feature.

Also, this is my first contribution working with PHPStan related code, and the progress in #397 looks pretty advanced to me (for now).

laytan: Do I understand correctly, that you would like to ask phpstan if the argument is of type 'string' or 'class-string' instead of this?

Do you have any idea where to start looking to come up with a solution that does that (because I am not really experienced in phpstan source)?

laytan: Also I saw #397 which if I am not wrong, will make this plugin obsolete?

If so, in my opinion, this PR would make for good initial support of the feature. Any further (cleaner) solutions, for example in #397, using the type system would then also be checked against the tests for this feature.

mglaman: We can ignore that other PR, it's experimental. Let's work on this first

mglaman (approved):

PR #436, opened 2022-06-23
Allow static::class . '::methodName' concatenation in lazy builders.

### LINKED ISSUE #371: Lazybuilder callback is expected wrongly to be array
Following error appears if you try to
`drupal-check -a web/core/modules/big_pipe/tests/modules/big_pipe_test/src/BigPipeTestController.php`

```
  71     #lazy_builder value 'non-empty-string' at key '0' is invalid.         
         💡 Refactor concatenation of `static::class` with method name to an    
         array callback: [static::class, 'counter']  
```

Array callback however does not work on these cases.

In `\Drupal\Core\Render\PlaceholderGenerator::createPlaceholder` the callback is expected to be in string format.
as given part of the build array is passed as is into
```
Html::escape($callback) 
```

longwave:
Discussed briefly in Slack with @mglaman - this would remove the false positives in https://www.drupal.org/project/drupal/issues/3259716 and allow `static::class . '::methodName'` to be resolved correctly.

Fixes #371

### DIFF
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -150,6 +150,12 @@ final class RenderCallbackRule implements Rule
                     ->tip('Change record: https://www.drupal.org/node/2966725.')
                     ->build();
             }
+
+            if (!$trustedCallbackType->isSuperTypeOf($type)->yes()) {
+                return RuleErrorBuilder::message(
+                    sprintf("%s callback class %s at key '%s' does not implement Drupal\Core\Security\TrustedCallbackInterface.", $keyChecked, $type->describe(VerbosityLevel::value()), $pos)
+                )->line($errorLine)->tip('Change record: https://www.drupal.org/node/2966725.')->build();
+            }
         } elseif ($type instanceof ConstantArrayType) {
             if (!$type->isCallable()->yes()) {
                 return RuleErrorBuilder::message(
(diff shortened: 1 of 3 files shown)


mglaman:
rebased off of `main` to fix failure

src/Rules/Drupal/RenderCallbackRule.php
```diff
@@ -217,7 +199,16 @@ private function doProcessNode(Node\Expr $node, Scope $scope, string $keyChecked
...
         $type = $scope->getType($node);
-        if ($type instanceof ConstantStringType) {
+        if ($type instanceof IntersectionType) {
+            // Covers concatenation of static::class . '::methodName'.
+            if ($node instanceof Node\Expr\BinaryOp\Concat) {
+                $leftType = $scope->getType($node->left);
+                $rightType = $scope->getType($node->right);
+                if ($leftType instanceof GenericClassStringType && $leftType->getGenericType() instanceof StaticType && $rightType instanceof ConstantStringType) {
+                    return new ConstantStringType($leftType->getGenericType()->getClassName() . $rightType->getValue());
+                }
```
mglaman, line 209: One problem we'll have is this in \PHPStan\Type\Constant\ConstantStringType::isCallable

```
        // 'MyClass::myStaticFunction'
        $matches = \_PHPStan_4b01b3801\Nette\Utils\Strings::match($this->value, '#^([a-zA-Z_\\x7f-\\xff\\\\][a-zA-Z0-9_\\x7f-\\xff\\\\]*)::([a-zA-Z_\\x7f-\\xff][a-zA-Z0-9_\\x7f-\\xff]*)\\z#');
        if ($matches !== null) {
            if (!$reflectionProvider->hasClass($matches[1])) {
                return \PHPStan\TrinaryLogic::createMaybe();
            }
```

Since we have a `static` keyword, the result will always be "maybe". That's how PHPStan works. The reflection provider will return `false` on `hasClass`.

mglaman (approved):

longwave:
LGTM :+1:

mglaman:
Thanks so much for your work!

andypost:
Great job! Looking for new release which could be added to https://www.drupal.org/project/drupal/issues/3293933

mglaman:
I was going to work on a few more items. But if releasing this one issue sooner than later unblocks core work, I can

Issue #438, opened 2022-06-30
entity_reference_revisions entity query access check false report


mglaman:
```
--------------------------------------------------------------------------------
Check manually 126  Missing explicit access check on entity query.              
--------------------------------------------------------------------------------
```

Code

```
    if ($composite_revision->isDefaultRevision()) {
      $count = $composite_storage
        ->getQuery()
        ->accessCheck(FALSE)
        ->allRevisions()
        ->condition($composite_storage->getEntityType()->getKey('id'), $composite_revision->id())
        ->count()
        ->execute();
      if ($count <= 1) {
        $composite_revision->delete();
        return TRUE;
      }
    }
```

PR #454, opened 2022-07-14
Entity query access check rule improvements

mglaman:
issue #438

### DIFF
--- a/tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
+++ b/tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
@@ -44,5 +44,10 @@ final class EntityQueryHasAccessCheckRuleTest extends DrupalRuleTestCase
                 ],
             ],
         ];
+
+        yield [
+            [__DIR__ . '/data/bug-438.php'],
+            []
+        ];
     }
 }
--- a/tests/src/Rules/data/bug-438.php
+++ b/tests/src/Rules/data/bug-438.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace EntityReferenceRevisionsExample;
+
+use Drupal\Core\Entity\ContentEntityInterface;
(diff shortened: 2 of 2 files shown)


tests/src/Rules/data/bug-437.php
```diff
@@ -0,0 +1,20 @@
...
+    public function __construct(EntityTypeManagerInterface $entityTypeManager)
+    {
+        $this->vocabularyQuery = $entityTypeManager
+            ->getStorage('taxonomy_term')
+            ->getQuery()
+            ->accessCheck(TRUE);
+    }
+    public function a(): int {
+        return $this->vocabularyQuery->count()->execute();
+    }
```
mglaman, line 19: Where this seems to break down:

```
            $returnType = new EntityQueryCountType(
                $varType->getClassName(),
                $varType->getSubtractedType(),
                $varType->getClassReflection()
            );
            if ($varType instanceof EntityQueryType && $varType->hasAccessCheck()) {
                return $returnType->withAccessCheck();
            }
```

In `\mglaman\PHPStanDrupal\Type\EntityQuery\EntityQueryDynamicReturnTypeExtension::getTypeFromMethodCall`. Previously it was set to be an access checked object type. But I believe the property typehint overrides what our dynamic extension return type has done.

tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
```diff
@@ -44,5 +44,40 @@ public function cases(): \Generator
...
+            ]
+        ];
+        yield 'bug-396a2.php' => [
+            [__DIR__ . '/data/bug-396a2.php'],
+            []
+        ];
+        yield 'bug-396a3.php' => [
+            [__DIR__ . '/data/bug-396a3.php'],
+            []
+        ];
```
mglaman, line 69: These pass when run individually, but fail when run in the same group.

PR #476, opened 2022-09-15
fix: Fix unpredictable false positive/negative of missing entity access check in function scope due to cache key conflict

### LINKED ISSUE #475: Random false positive "Missing explicit access check on entity query"
I've been hitting some issues for a month or so where randomly after switching a branch I would see several "Missing explicit access check on entity query" reported. However, they're all fine and fixed. I would move the accessCheck() call around and then run again and the error would vanish. I would put the line back to the original and it wouldn't fail anymore it would be like a ghost that's now gone.

Push forward to now, and we upgraded PHPStan and went from 1.1.20 phpstan-drupal to 1.1.25 and now we have 6 false positives in our CI which won't go away. No matter what we do with the accessCheck call they report. And if we checkout the exact same code version and run PHPStan we get zero errors. So we seem to be seeing different results depending on where it runs - somewhat similar to before where random people would get random failures and it would just disappear - but instead it's now on the CI, no longer random, and completely reproducing every time. CI is GitHub Actions hosted so ephemeral. We just can't get the PHPStan to pass anymore. Locally, we get 0 errors and cannot get them to appear. I managed to get them once but they vanished on the second run, but no luck since.

Can't rule out something really weird in our configuration but I've yet to find it. The only difference I haven't explored in detail is GitHub Actions is using Ubuntu and I am using macOS.

It's only ever been a problem with this specific error though.

driskell:
Fixes #475

The EntityQueryType was getting cached in function scope for a `condition` call and because it wasn't handled by dynamic return it reused. So if you had `condition` before `accessCheck` and then later on after it, the latter one could fail because it would return the cached without access check and there would be no subsequent access check (it happens before).

Added test case.

Still examining why this doesn't occur in class scope but it seems phpstan has additional cache key when it is just not sure of its granularity.

### DIFF
--- a/src/Type/EntityQuery/EntityQueryType.php
+++ b/src/Type/EntityQuery/EntityQueryType.php
@@ -17,14 +17,17 @@ class EntityQueryType extends ObjectType
 
     public function withAccessCheck(): self
     {
-        $type = clone $this;
+        $type = new self(
+            $this->getClassName(),
+            $this->getSubtractedType(),
+            $this->getClassReflection()
+        );
         $type->hasAccessCheck = true;
-
         return $type;
     }
 
     protected function describeAdditionalCacheKey(): string
     {
-        return $this->hasAccessCheck ? 'with-access-check' : '';
+        return $this->hasAccessCheck ? 'with-access-check' : 'without-access-check';
     }
 }
--- a/src/Type/EntityQuery/EntityQueryType.php
+++ b/src/Type/EntityQuery/EntityQueryType.php
@@ -17,7 +17,10 @@ class EntityQueryType extends ObjectType
 
     public function withAccessCheck(): self
     {
-        $type = new self(
+        // The constructor of ObjectType is under backward compatibility promise.
+        // @see https://phpstan.org/developing-extensions/backward-compatibility-promise
+        // @phpstan-ignore-next-line
+        $type = new static(
             $this->getClassName(),
             $this->getSubtractedType(),
             $this->getClassReflection()
--- a/src/Type/EntityQuery/EntityQueryType.php
+++ b/src/Type/EntityQuery/EntityQueryType.php
@@ -22,4 +22,9 @@ class EntityQueryType extends ObjectType
 
         return $type;
     }
+
+    protected function describeAdditionalCacheKey(): string
+    {
+        return $this->hasAccessCheck ? 'with-access-check' : '';
+    }
 }
--- a/tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
+++ b/tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
@@ -86,5 +86,15 @@ final class EntityQueryHasAccessCheckRuleTest extends DrupalRuleTestCase
             [__DIR__ . '/data/bug-437.php'],
             []
         ];*/
+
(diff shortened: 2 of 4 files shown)


driskell:
It did happen in class scope - PR has tests for both that reproduce and the fix solve

mxr576:
Tested the fix on those false positives that we got reported and the patch did silence them. RTBC

(The pipeline looks broken at this moment, I made an attempt to fix one issue in #485, but other skeletons popped up.)

src/Type/EntityQuery/EntityQueryType.php
```diff
@@ -22,4 +22,9 @@ public function withAccessCheck(): self
 
         return $type;
     }
+
+    protected function describeAdditionalCacheKey(): string
+    {
+        return $this->hasAccessCheck ? 'with-access-check' : '';
+    }
```
mglaman, line 29: Should we make sure that we has `without-access-check` as the other option for verbosity?

tests/src/Rules/EntityQueryHasAccessCheckRuleTest.php
```diff
@@ -86,5 +86,15 @@ public function cases(): \Generator
...
+
+        yield 'bug-475.php' => [
+            [__DIR__.'/data/bug-475.php'],
+            []
+        ];
+
+        yield 'bug-475b.php' => [
+            [__DIR__.'/data/bug-475b.php'],
+            []
+        ];
```
mglaman, line 98: The tests are failing on these:

```
'17: Missing explicit access check on entity query.
    💡 See https://www.drupal.org/node/3201242
```

mglaman (commented):
I think I found the problem. It is in `\mglaman\PHPStanDrupal\Type\EntityQuery\EntityQueryType::withAccessCheck` because we clone the object, which has a cached description key. This is definitely a fix, but we need to stop cloning, as it copies the `cachedDescription`

src/Type/EntityQuery/EntityQueryType.php
```diff
@@ -17,9 +17,17 @@ public function hasAccessCheck(): bool
 
     public function withAccessCheck(): self
     {
-        $type = clone $this;
+        $type = new self(
+            $this->getClassName(),
+            $this->getSubtractedType(),
+            $this->getClassReflection()
+        );
         $type->hasAccessCheck = true;
```
mglaman, line 25: cc @brambaud since you contributed this code. I didn't think twice when we first added this, but cloning was copying over cached information, leading to the bugs.

mglaman: :/ looks like this caused more failures although it fixed the one bug. Probably because I used `self` instead of `static`.

driskell:
@mglaman Thanks! I read the two commits and they definitely improvements :)

Really happy this got in - it was a super hard one to track down!!!

mglaman:
> Really happy this got in - it was a super hard one to track down!!!

Thanks for tracking this down! It was driving me wild.

src/Type/EntityQuery/EntityQueryType.php
brambaud: Hooo! 😮 
Nice catch! Didn't think of that! 🙇

PR #495, opened 2023-01-04
False negative accessCheck failure on entity queries with entity type coming from parameter 

### LINKED ISSUE #494: False negative checkAccess failure on count entity queries with entity type coming from parameter
Almost all problems with the `checkAccess` on entity queries are gone with #476.
Whilst working on https://www.drupal.org/project/drupal/issues/3330981 I've found a few escapees.

Boegie:
Fixes #494.

### DIFF
--- a/src/Type/DrupalStaticEntityQueryDynamicReturnTypeExtension.php
+++ b/src/Type/DrupalStaticEntityQueryDynamicReturnTypeExtension.php
@@ -57,8 +57,13 @@ class DrupalStaticEntityQueryDynamicReturnTypeExtension implements DynamicStatic
         if ($type instanceof ConstantStringType) {
             $entityTypeId = $type->getValue();
         } else {
-            // @todo determine what these types are, and try to resolve entity name from.
-            return $returnType;
+            // We're unsure what specific EntityQueryType it is, so let's stick
+            // with the general class itself, so at least it gets access checked
+            return new EntityQueryType(
+                $returnType->getClassName(),
+                $returnType->getSubtractedType(),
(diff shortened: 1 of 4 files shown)


Boegie:
Ok, looks like the CI is setup in a way that if one job fails, all others are cancelled.

Anyway: `PHP 7.4 | Drupal ^8.9` shows the same test failures I get on `10.1.x-dev`:
```
1) mglaman\PHPStanDrupal\Tests\Rules\EntityQueryHasAccessCheckRuleTest::test with data set "bug-494.php" (array('/home/runner/work/phpstan-dru...94.php'), array())
--- Expected
+++ Actual
@@ @@
-'
+'09: Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked.
+    💡 See https://www.drupal.org/node/3201242
+15: Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked.
+    💡 See https://www.drupal.org/node/3201242
```

Boegie:
Added a fix(?), works for the new test cases, doesn't break other test cases. 

Ran these changes on Drupal Core and false negatives disappeared and new (seemingly) legit warnings appeared.

mglaman:
Thanks @Boegie !

Issue #496, opened 2023-01-04
False negative accessCheck failure on non-chained entity queries


Boegie:
Almost all problems with the `accessCheck` on entity queries are gone with #476.
Whilst working on https://www.drupal.org/project/drupal/issues/3330981 I've found a few escapees.

This one is about false negative `accessCheck` failure on entity queries that are _not_ chained.

I've added a test-case in the attached PR, it's probably more easily explained in code (and test failures) than words.

mglaman:
Sounds like the opposite problem https://github.com/mglaman/phpstan-drupal/issues/388 was opened for, where extreme chaining things broke (I think.)

mglaman:
Opened a discussion https://github.com/phpstan/phpstan/discussions/8773

I don't think this is possible because we cannot track side effects on objects.

mglaman:
I am silly; https://phpstan.org/developing-extensions/type-specifying-extensions is the answer after all. I assumed not because I have only seen implementations that read the args.

mglaman:
This was fixed with https://github.com/mglaman/phpstan-drupal/pull/497

PR #497, opened 2023-01-04
False negative accessCheck failure on non-chained entity queries

Boegie:
Testcases for #496.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -297,3 +297,7 @@ services:
 	-
 		class: mglaman\PHPStanDrupal\Drupal\DrupalStubFilesExtension
 		tags: [phpstan.stubFilesExtension]
+	-
+		class: mglaman\PHPStanDrupal\Type\EntityQuery\AccessCheckTypeSpecifyingExtension
+		tags:
+			- phpstan.typeSpecifier.methodTypeSpecifyingExtension
--- a/src/Type/EntityQuery/AccessCheckTypeSpecifyingExtension.php
+++ b/src/Type/EntityQuery/AccessCheckTypeSpecifyingExtension.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Type\EntityQuery;
+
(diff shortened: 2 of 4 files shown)


Boegie:
I would expect no failures here.

Instead I get:

```
There was 1 failure:

1) mglaman\PHPStanDrupal\Tests\Rules\EntityQueryHasAccessCheckRuleTest::test with data set "bug-496.php" (array('/home/runner/work/phpstan-dru...96.php'), array())
--- Expected
+++ Actual
@@ @@
-'
+'15: Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked.
+    💡 See https://www.drupal.org/node/3201242
+20: Relying on entity queries to check access by default is deprecated in drupal:9.2.0 and an error will be thrown from drupal:10.0.0. Call \Drupal\Core\Entity\Query\QueryInterface::accessCheck() with TRUE or FALSE to specify whether access should be checked.
+    💡 See https://www.drupal.org/node/3201242
```

Note: Both entity queries pass when used chained. so without the `$query`-bit on each line.

mglaman:
It breaks down due to PHPStan not tracking the side effect of calling `accessCheck`, which modifies the state of the query.

In `\PHPStan\Analyser\NodeScopeResolver::processStmtNode` we have:

```
        } elseif ($stmt instanceof Node\Stmt\Expression) {
            $earlyTerminationExpr = $this->findEarlyTerminatingExpr($stmt->expr, $scope);
            $result = $this->processExprNode($stmt->expr, $scope, $nodeCallback, \PHPStan\Analyser\ExpressionContext::createTopLevel());
```

Screenshot 2023-01-25 at 3 17 31 PM

`$query` is in its initial state of not being flagged. But `$query->accessCheck(FALSE);` has itself considered as returning the query with access check. But this does not update the existing `$query` resolved type because it's a side effect not a return typ.

PR #498, opened 2023-01-05
474

### LINKED ISSUE #474: getAggregateQuery() entity query access check false report
Probably a follow-up to the improvements that were made in https://github.com/mglaman/phpstan-drupal/pull/454...

The following code is resulting in PHPStan error: "Missing explicit access check on entity query."

```
$query = $this->entityTypeManager->getStorage($entity_type)->getAggregateQuery()
  ->accessCheck(TRUE)
  ->condition('type', $bundle);
```

Perhaps we just need to accommodate `getAggregateQuery()`? Or is there a "correct" way to structure the code to make PHPStan happy? :-)

Boegie:
Fixes #474


Boegie:
So with the added test cases for this issue, I would expect 0 test failures.

PR #523, opened 2023-03-30
Make EntityFieldReflection aware of the book property added by book.module.

mparker17:
# Problem / motivation

As of Drupal 9.5, 10.0, and 10.1, the book module adds a "book" property to Node instances that have been enabled as books.

PHPStan is not aware of this and thus reports messages like `Property Drupal\node\NodeInterface::$book (Drupal\Core\Field\FieldItemListInterface) does not accept array.`

# Proposed resolution

Tell PHPStan to expect a `book` property on Nodes that is an array.

# Additional notes

I wanted to be more specific about what's in this array, but I'm not certain how to do that in PHPStan - its API documentation is somewhat minimal. If you have suggestions, I would be happy to update this merge request.

The `book` property is an array that contains the following keys/types:

| Array key | Data type |
|--------|--------|
| `nid` | `int` (or `numeric-string`) |
| `bid` | `int` (or `numeric-string`) |
| `pid` | `int` (or `numeric-string`) |
| `has_children` | `int`  (or `numeric-string` or `bool`) |
| `weight` | `int` (or `numeric-string`) | 
| `depth` | `int` (or `numeric-string`) | 
| `p1` | `int` (or `numeric-string`) | 
| `p2` | `int` (or `numeric-string`) | 
| `p3` | `int` (or `numeric-string`) | 
| `p4` | `int` (or `numeric-string`) | 
| `p5` | `int` (or `numeric-string`) | 
| `p6` | `int` (or `numeric-string`) | 
| `p7` | `int` (or `numeric-string`) | 
| `p8` | `int` (or `numeric-string`) | 
| `p9` | `int` (or `numeric-string`) | 
| `link_path` | `string` |
| `link_title` | `string` |

... See [book_schema() in 10.1.x core for more info](https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/book/book.install#L23-126). While many of these fields are stored as integers in the database, the book module usually queries them and dumps the query result into the array, and due to the way the database layer works, they come out as strings containing the integers that were stored in the database.

### DIFF
--- a/src/Reflection/EntityFieldReflection.php
+++ b/src/Reflection/EntityFieldReflection.php
@@ -4,8 +4,10 @@ namespace mglaman\PHPStanDrupal\Reflection;
 
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Type\ArrayType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 
 /**
@@ -41,6 +43,12 @@ class EntityFieldReflection implements PropertyReflection
             return new ObjectType($objectType);
         }
 
+        if ($this->propertyName === 'book'
+            && $this->declaringClass->is('Drupal\node\NodeInterface')
+        ) {
+            return new ArrayType(new StringType(), new MixedType());
+        }
+
         if ($this->declaringClass->isSubclassOf('Drupal\Core\Entity\ContentEntityInterface')) {
             // Assume the property is a field.
             return new ObjectType('Drupal\Core\Field\FieldItemListInterface');
@@ -60,6 +68,12 @@ class EntityFieldReflection implements PropertyReflection
                 $objectType = 'Drupal\Core\Entity\EntityInterface';
             }
             return new ObjectType($objectType);
+        }
+
+        if ($this->propertyName === 'book'
+            && $this->declaringClass->is('Drupal\node\NodeInterface')
+        ) {
+            return new ArrayType(new StringType(), new MixedType());
         }
 
         // @todo Drupal allows $entity->field_myfield = 'string'; does this break that?
--- a/src/Reflection/EntityFieldReflection.php
+++ b/src/Reflection/EntityFieldReflection.php
@@ -4,10 +4,8 @@ namespace mglaman\PHPStanDrupal\Reflection;
 
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertyReflection;
-use PHPStan\Type\ArrayType;
 use PHPStan\Type\MixedType;
 use PHPStan\Type\ObjectType;
-use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
 
 /**
--- a/src/Reflection/EntityFieldReflection.php
+++ b/src/Reflection/EntityFieldReflection.php
@@ -43,12 +43,6 @@ class EntityFieldReflection implements PropertyReflection
             return new ObjectType($objectType);
         }
 
-        if ($this->propertyName === 'book'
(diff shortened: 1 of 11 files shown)


mparker17:
At time-of-writing, the CircleCI Pipeline is failing with "Could not find a usable config.yml, you may have revoked the CircleCI OAuth app." - my gut says this is an upstream issue?

mglaman:
CircleCI - thanks for heads up, I'll need to see why that broke.

Book module - I'm guessing this isn't a computed base field but a generic property set like `original`. Based on a quick mobile review this makes sense. There isn't a better way to handle it, honestly.

But we could possibly also fix this via a stub which uses PHPStan's array phpDoc syntax.

Something like


```
@property array{foo: string} $book
```

mparker17:
> You're right. I think another issue captured these reflection helpers override stubs

:+1: 

****

Forgive my confusion, but I'm not sure what you'd like me to do next (if anything)! Thanks for your patience with me!

mglaman:
Oh nothing else needs to be done. I'll review the PR this week

src/Reflection/EntityFieldReflection.php
```diff
@@ -41,6 +43,12 @@ public function getReadableType(): Type
             return new ObjectType($objectType);
         }
 
+        if ($this->propertyName === 'book'
+            && $this->declaringClass->is('Drupal\node\NodeInterface')
```
mglaman, line 47: Should this be `isSubclassOf`?

src/Reflection/EntityFieldReflection.php
```diff
@@ -62,6 +70,12 @@ public function getWritableType(): Type
...
```
mglaman, line 74: Same: should this be `isSubclassOf`?

mglaman (changes requested):
I think we need `isSubclassOf` so that we can account for concrete class of `Node` and bundle classes.

Also, we should add a test. I can work on that if you'd like.

src/Reflection/EntityFieldReflection.php
mparker17: @mglaman sorry for the delay in answering.

I actually tried this... and it didn't work (for my code anyway).

I'll try to explain as best I can why that is, but be aware (a) it was a few weeks ago so the details are a bit fuzzy and (b) I'm not an expert in how PHPStan understands things.

We had some functions that accepted NodeInterface as input, i.e.: `function foo(NodeInterface $a)` and others that accepted Node as input, i.e.: `function bar(Node $b)`.

Inside `function bar(Node $b)`, if we set `$b->bid = 123`, then `$this->declaringClass->isSubclassOf(...)` would return TRUE, i.e.: work just fine. But inside `function foo(NodeInterface $a)`, if we set `$a->bid = 123` then ``$this->declaringClass->isSubclassOf(...)` would return FALSE, i.e.: fail.

A bit of debugging suggests that `$this->declaringClass` would show up as a `NodeInterface`, and `isSubclassOf` would fail, I think this is because `\PHPStan\BetterReflection\Reflection\Adapter\ReflectionClass::isSubclassOf()` calls `getParentClassNames()` and it doesn't show up in that list.

Anyway, after much debugging, it seemed like `$this->declaringClass->is()` worked fine in both cases.

src/Reflection/EntityFieldReflection.php
mparker17: (same answer as above)

mglaman:
Sorry! I forgot about this, I'll make my way back to it this week.

mglaman:
I have one concern: there are discussions on moving Book to contrib; see https://www.drupal.org/project/ideas/issues/3265493.

This extension would then hold on to custom property reflections for just one contrib when many may manipulate node entities this way. Considering this while giving second review

mglaman:
Okay! I fixed the issue of parsing `@property` on entity stubs, so I moved the book property types to the stub which seemed less hacky to maintain and makes it easier to eventually remove any funky entity field reflection code

Issue #516, opened 2023-02-28
False positives from loadInclude rule


alexpott:
Code like
```
  $module_handler = \Drupal::moduleHandler();
  $enabled_modules = $module_handler->getModuleList();

  foreach (array_keys($enabled_modules) as $module) {
    // All modules should have a recorded schema version, but when they
    // don't, detect and fix the problem.
    if (!isset($versions[$module])) {
      // Ensure the .install file is loaded.
      $module_handler->loadInclude($module, 'install');
      $all_updates = $update_registry->getAvailableUpdates($module);
```
(see update.inc)

results in 
```
		-
			message: "#^A file could not be loaded from Drupal\\\\Core\\\\Extension\\\\ModuleHandlerInterface\\:\\:loadInclude$#"
			count: 1
			path: includes/update.inc
```
which it should not. There are instances where the first argument to loadInclude is a variable and in these instance PHPStan should not result in an error.

PR #524, opened 2023-04-07
Ignore LoadIncludes rule if module and file name unknown

Boegie:
Fixes https://github.com/mglaman/phpstan-drupal/issues/516

### DIFF
--- a/src/Rules/Drupal/LoadIncludes.php
+++ b/src/Rules/Drupal/LoadIncludes.php
@@ -48,7 +48,9 @@ class LoadIncludes extends LoadIncludeBase
         try {
             // Try to invoke it similarly as the module handler itself.
             [$moduleName, $filename] = $this->parseLoadIncludeArgs($args[0], $args[1], $args[2] ?? null, $scope);
-            if ($moduleName === null && $filename == null) {
+            if (!$moduleName && !$filename) {
+                // Couldn't determine module- nor file-name, most probably
+                // because it's a variable. Nothing to load, bail now.
                 return [];
             }
             $module = $this->extensionMap->getModule($moduleName);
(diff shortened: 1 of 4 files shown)


src/Rules/Drupal/LoadIncludes.php
```diff
@@ -48,6 +48,9 @@ public function processNode(Node $node, Scope $scope): array
         try {
             // Try to invoke it similarly as the module handler itself.
             [$moduleName, $filename] = $this->parseLoadIncludeArgs($args[0], $args[1], $args[2] ?? null, $scope);
+            if ($moduleName === null && $filename == null) {
+                return [];
+            }
```
Boegie, line 53: $moduleName could not be found (because it's a variable), proceeding to `$module = $this->extensionMap->getModule($moduleName);` would throw a `TypeError` with message "mglaman\PHPStanDrupal\Drupal\ExtensionMap::getModule(): Argument #1 ($name) must be of type string, null given, called in D:\git\phpstan-drupal\src\Rules\Drupal\LoadIncludes.php on line 54".

That would be caught by the `catch` later on in this class, which results in the error "A file could not be loaded from Drupal\Core\Extension\ModuleHandlerInterface::loadInclude", which is creating a false positive.

Bailing here makes it all better.

mglaman: Should `parseLoadIncludeArgs` be changed to not just return `[]` but an array of the same size each time when something is `null`?

This makes sense to just bail out.

Boegie: So something like my latest commit?

Boegie: Sense = made :)

mglaman (approved):
Looks good. Reviewed via mobile, I'll merge when I can get to my laptop and give it one more look over

PR #529, opened 2023-04-23
Support TrustedCallback attribute for RenderCallbackRule

### LINKED ISSUE #527: Add support for TrustedCallback attribute
https://www.drupal.org/node/3274867 added the possibility of using the `TrustedCallback` on a method instead of having the whole class `...implements TrustedCallbackInterface`.

If new code uses this method `phpstan` checks fail like this:
```
41     #lazy_builder callback class 'Drupal\announcements_feed\LazyBuilders'  
         at key '0' does not implement                                          
         Drupal\Core\Security\TrustedCallbackInterface.                         
         💡 Change record: https://www.drupal.org/node/2966725.
```
(can be seen here https://www.drupal.org/pift-ci-job/2646296)

When debugging locally, Drupal code is happy with the new attributes.

The affected file is mostly https://github.com/mglaman/phpstan-drupal/blob/main/src/Rules/Drupal/RenderCallbackRule.php

mglaman:
Fixes #527

### DIFF
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -208,7 +208,7 @@ final class RenderCallbackRule implements Rule
                                 $typeAndMethodName->getType()->describe(VerbosityLevel::value()),
                                 $pos
                             )
-                        )->line($errorLine)->tip('Change record: https://www.drupal.org/node/2966725')->build();
+                        )->line($errorLine)->tip('Change record: https://www.drupal.org/node/2966725.')->build();
                     }
                 }
             }
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -149,6 +149,9 @@ final class RenderCallbackRule implements Rule
                     ->build();
             }
 
+            $hasTrustedCallbackAttribute = $type->getCallableParametersAcceptors($scope);
+            $isTrustedCallbackType = $trustedCallbackType->isSuperTypeOf($type)->yes();
+
             if (!$trustedCallbackType->isSuperTypeOf($type)->yes()) {
                 return RuleErrorBuilder::message(
                     sprintf("%s callback class %s at key '%s' does not implement Drupal\Core\Security\TrustedCallbackInterface.", $keyChecked, $type->describe(VerbosityLevel::value()), $pos)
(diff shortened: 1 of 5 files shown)


tests/src/Rules/data/bug-527.php
```diff
@@ -0,0 +1,23 @@
...
+    public function render(): array {
+        return [
+            'one' => [
+                '#lazy_builder' => [
+                    [self::class, 'someCallback'],
+                    ['bar']
+                ]
+            ]
+        ];
+    }
```
mglaman, line 22: This needs to test:

- pre_render
- post_render
- access_callback

src/Rules/Drupal/RenderCallbackRule.php
```diff
@@ -166,7 +169,19 @@ private function doProcessNode(Node\Expr $node, Scope $scope, string $keyChecked
...
+                } else {
+                    // @todo does PHPStan have a better way?
+                    // @see https://github.com/phpstan/phpstan/discussions/5863
+                    $methodNativeClassReflection = $testReflection->getDeclaringClass()->getNativeReflection();
+                    $methodNativeReflection = $methodNativeClassReflection->getMethod($typeAndMethodName->getMethod());
+                    $isTrustedMethod = (bool) $methodNativeReflection->getAttributes('Drupal\Core\Security\Attribute\TrustedCallback');
+                }
+
+                $isTrustedCallbackType = $trustedCallbackType->isSuperTypeOf($typeAndMethodName->getType())->yes();
+                if (!$isTrustedCallbackType && !$isTrustedMethod) {
```
mglaman, line 184: This needs to be cleaned up. The discussion is from 2021, but I don't see any "better" way to reflect the method and get its attributes. Also, this should only count on Drupal >= 10.1.x. Otherwise 10.0 will see a false report about being able to use the attribute.

src/Rules/Drupal/RenderCallbackRule.php
```diff
@@ -149,6 +149,9 @@ private function doProcessNode(Node\Expr $node, Scope $scope, string $keyChecked
                     ->build();
             }
 
+            $hasTrustedCallbackAttribute = $type->getCallableParametersAcceptors($scope);
+            $isTrustedCallbackType = $trustedCallbackType->isSuperTypeOf($type)->yes();
+
```
mglaman, line 154: remove

mglaman (commented):
This is a proof of concept fix. Needs work. Only fixes array callables.

PR #531, opened 2023-04-24
Count entity queries on config entities incorrectly require accesCheck()

### LINKED ISSUE #530: Count entity queries on config entities incorrectly require accesCheck()
When ever `phpstan-drupal` encounters a count entity query on config entities, it requires them to have an `accesCheck()`, which is incorrect.

"Normal" entity queries on config entities correctly _don't_ require the `accesCheck()`

Boegie:
resolve #530


Boegie:
Might be taking (unwanted) shortcuts, but it passes tests and seems to work.

Boegie:
Agreed, as said, I _was_ using shortcuts...

There are more issues with the count entity queries, will try to add test cases here and dive in when I have the time.

PR #544, opened 2023-04-28
Support concatenated callables in RenderCallbackRule

### LINKED ISSUE #543: Concatenated callables treated as union type, used to be skipped on RenderCallbackRule
# Bug report

Created by #529

Discovered in https://www.drupal.org/project/drupal/issues/3357114.

`\Drupal\Tests\Core\Render\RendererTest::testRenderWithAccessControllerResolved` used to pass, but not any more.

```
+		-
+			message: "#^\\#access_callback callback class 'Drupal\\\\\\\\Tests\\\\\\\\Coreâ€¦' at key '0' does not implement Drupal\\\\Core\\\\Security\\\\TrustedCallbackInterface\\.$#"
+			count: 4
+			path: tests/Drupal/Tests/Core/Render/RendererTest.php
```

"Interestingly", when replacing the variable with all possible strings, it still passes.

See test cases in PR #544.

Boegie:
Will resolve #543

### DIFF
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -153,7 +153,7 @@ final class RenderCallbackRule implements Rule
                 )->line($errorLine)
                     ->tip('Change record: https://www.drupal.org/node/2966725.')
                     ->build();
-            } elseif (!$trustedCallbackType->isSuperTypeOf($type)->yes()) {
+            } elseif (!$trustedCallbackType->isSuperTypeOf(new ObjectType(explode('::', $constantStringType->getValue())[0]))->yes()) {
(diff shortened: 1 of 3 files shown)


Boegie:
Originally, being a `UnionType`, the callback ended up here: https://github.com/mglaman/phpstan-drupal/blob/ed32c286a2e72e2e8919f13ef8faabad00ee105b/src/Rules/Drupal/RenderCallbackRule.php#L194, so rather unchecked.

Since we have an array of  `ConstantStringType`, I've tried to check it.

mglaman:
Ah thanks, I overlooked this test case in the fix. I'll review when at my laptop and make a fix ASAP.

mglaman (approved):

PR #558, opened 2023-05-12
 False positive when (one of) the callback function name(s) in #date_date_callbacks/#date_time_callbacks is a variable. 

### LINKED ISSUE #553: False positive when (one of) the callback function name(s) in `#date_date_callbacks`/`#date_time_callbacks` is a variable.
# Bug report
_(Yes, it's me again...)_

With #551 we introduced a false positive when (one of) the callback function name(s) in `#date_date_callbacks`/`#date_time_callbacks` is a variable.

Code taken from `\Drupal\KernelTests\Core\Datetime\DatelistElementFormTest`

```
  public function datelistDateCallbackTrusted(array &$element, FormStateInterface $form_state, DrupalDateTime $date = NULL) {
    $element['datelistDateCallbackExecuted'] = [
      '#value' => TRUE,
    ];
    $form_state->set('datelistDateCallbackExecuted', TRUE);
  }

  public function buildForm(array $form, FormStateInterface $form_state, string $date_callback = 'datelistDateCallbackTrusted') {

    $form['datelist_element'] = [
      '#title' => 'datelist test',
      '#type' => 'datelist',
      '#default_value' => new DrupalDateTime('2000-01-01 00:00:00'),
      '#date_part_order' => [
        'month',
        'day',
        'year',
        'hour',
        'minute', 'ampm',
      ],
      '#date_text_parts' => ['year'],
      '#date_year_range' => '2010:2020',
      '#date_increment' => 15,
      '#date_date_callbacks' => [[$this, $date_callback]],
    ];

    $form['submit'] = [
      '#type' => 'submit',
      '#value' => t('Submit'),
    ];

    return $form;
  }
```

returns:

```
		-
			message: "#^\\#date_date_callbacks callback array\\{\\$this\\(Drupal\\\\KernelTests\\\\Core\\\\Datetime\\\\DatelistElementFormTest\\), string\\} at key '0' is not callable\\.$#"
			count: 1
			path: tests/Drupal/KernelTests/Core/Datetime/DatelistElementFormTest.php
```

Boegie:
Tries to resolve #553

### DIFF
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -174,12 +174,19 @@ final class RenderCallbackRule implements Rule
 
         foreach ($type->getConstantArrays() as $constantArrayType) {
             if (!$constantArrayType->isCallable()->yes()) {
-                if ($constantArrayType->getItemType() instanceof UnionType
-                    && !($constantArrayType->getValuesArray()->getValueTypes()[1] instanceof ConstantStringType)) {
-                    // Right-hand side of UnionType is a variable. Nothing to
-                    // check, bail now.
-                    $checkIsCallable = false;
-                    break;
+                // If the right-hand side of the array is a variable, we cannot
+                // determine if it is callable. Bail now.
+                $itemType = $constantArrayType->getItemType();
+                if ($itemType instanceof UnionType) {
+                    $unionConstantStrings = array_merge(...array_map(static function (Type $type) {
+                        return $type->getConstantStrings();
+                    }, $itemType->getTypes()));
+                    if (count($unionConstantStrings) === 0) {
+                        // Right-hand side of UnionType is not a constant string. We cannot determine if the dynamic
+                        // value is callable or not.
+                        $checkIsCallable = false;
+                        break;
+                    }
                 }
                 $errors[] = RuleErrorBuilder::message(
                     sprintf("%s callback %s at key '%s' is not callable.", $keyChecked, $constantArrayType->describe(VerbosityLevel::value()), $pos)
--- a/src/Rules/Drupal/RenderCallbackRule.php
+++ b/src/Rules/Drupal/RenderCallbackRule.php
@@ -28,7 +28,7 @@ use PHPStan\Type\Type;
 use PHPStan\Type\UnionType;
 use PHPStan\Type\VerbosityLevel;
 
-final class TrustedCallbackRule implements Rule
+final class RenderCallbackRule implements Rule
 {
 
     private ReflectionProvider $reflectionProvider;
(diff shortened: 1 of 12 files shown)


src/Rules/Drupal/TrustedCallbackRule.php
```diff
@@ -172,6 +174,13 @@ private function doProcessNode(Node\Expr $node, Scope $scope, string $keyChecked
 
         foreach ($type->getConstantArrays() as $constantArrayType) {
             if (!$constantArrayType->isCallable()->yes()) {
+                if ($constantArrayType->getItemType() instanceof UnionType
+                    && !($constantArrayType->getValuesArray()->getValueTypes()[1] instanceof ConstantStringType)) {
```
Boegie, line 178: This makes my eyes bleed, but couldn't find a decent way to determine if the right hand side of the UnionType is not a ConstantStringType, but a StringType, which (in my undestandi9ng of PHPStan magic) makes this a variable.

Boegie:
Failing check is the new baseline, with correctly removed suppressions.

mglaman (changes requested):
Renaming the class would be a breaking change as folks may have manually configured all of their rules (I don't know who would have, but I've been saving renaming/splitting this out for 1.2.x.)

Boegie:
It seems one is never free of The Chains Of BC, even in here...

I'll revert the renaming.

mglaman:
I opened https://github.com/mglaman/phpstan-drupal/issues/560 to take action on this, finally

src/Rules/Drupal/RenderCallbackRule.php
```diff
@@ -172,6 +174,13 @@ private function doProcessNode(Node\Expr $node, Scope $scope, string $keyChecked
 
         foreach ($type->getConstantArrays() as $constantArrayType) {
             if (!$constantArrayType->isCallable()->yes()) {
+                if ($constantArrayType->getItemType() instanceof UnionType
+                    && !($constantArrayType->getValuesArray()->getValueTypes()[1] instanceof ConstantStringType)) {
```
Boegie, line 178: This makes my eyes bleed, but couldn't find a decent way to determine if the right hand side of the UnionType is not a ConstantStringType, but a StringType, which (in my undestandi9ng of PHPStan magic) makes this a variable.

mglaman: I think I found a way! Took a bit though.

mglaman (approved):

src/Rules/Drupal/RenderCallbackRule.php
Boegie: Nice, not as "grokable" as expected, but that's what's the PHPStan issue is for.

PR #562, opened 2023-05-12
PR for #561: RenderCallbackRule ignores ::trustedCallbacks()

### LINKED ISSUE #561: `RenderCallbackRule` ignores `::trustedCallbacks()`
# Bug report
_(Now that all Entity Query problems seem to be solved, I looked a bit closer at the Trusted Callbacks, so expect some more issues on that)_

Currently `\mglaman\PHPStanDrupal\Rules\Drupal\RenderCallbackRule` only checks if the class implements `TrustedCallbackInterface`.
However the callbacks should also be in the return array from `trustedCallbacks()` to be trusted.

Now I _highly_ doubt that people "in the wild" go through the trouble of creating callbacks and making sure the containing class implements `TrustedCallbackInterface` and finally _don't_ add the callback to `trustedCallbacks()`, so the impact seems low.

However what we're currently doing seems incorrect.

See testcases in (extended) `tests/src/Rules/data/bug-543.php`

Boegie:
Resolves #561

### DIFF
--- a/tests/src/Rules/RenderCallbackRuleTest.php
+++ b/tests/src/Rules/RenderCallbackRuleTest.php
@@ -161,7 +161,16 @@ final class RenderCallbackRuleTest extends DrupalRuleTestCase {
 
         yield [
             __DIR__ . '/data/bug-543.php',
-            []
+            [
+                [
+                    "#access_callback callback array{static(Bug543\TestAccessClass), 'accessResultForbiddenNotInTrustedCallbacks'} at key '0' does not implement Drupal\Core\Security\TrustedCallbackInterface.",
+                    58
+                ],
+                [
+                    "#access_callback callback array{static(Bug543\TestAccessClass), 'accessResultForbiddenNotInTrustedCallbacks'} at key '0' does not implement Drupal\Core\Security\TrustedCallbackInterface.",
+                    103
+                ],
+            ],
         ];
     }
 
--- a/tests/src/Rules/data/bug-543.php
+++ b/tests/src/Rules/data/bug-543.php
@@ -21,10 +21,13 @@ class TestClass {
             [TRUE],
             [AccessResult::forbidden()],
             [AccessResult::allowed()],
+            ['accessResultForbiddenNotInTrustedCallbacks'],
         ];
     }
 
     /**
+     * Tests callbacks with the method names in a variable.
+     *
      * @dataProvider providerAccessValues
      */
     public function testRenderWithAccessControllerResolved($access) {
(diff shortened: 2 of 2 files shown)


mglaman:
😩 

```
     Internal error: Internal error: Method of class                                                  
     Drupal\Tests\Core\Render\PlaceholdersTest cannot be used as the class                            
     does not exist in file                                                                           
     /home/runner/work/_temp/drupal/core/tests/Drupal/Tests/Core/Render/PlaceholderGeneratorTest.php  
```

That is because `PlaceholdersTest` is in `core/tests/Drupal/Tests/Core/Render/RendererTestBase.php`. The test class `\Drupal\Tests\Core\Render\PlaceholderGeneratorTest` extends `\Drupal\Tests\Core\Render\RendererTestBase`. But it's not brought into scope.

src/Drupal/DrupalAutoloader.php
```diff
@@ -221,6 +221,14 @@ protected function addCoreTestNamespaces(): void
...
         $this->namespaces['Drupal\\Tests\\TestSuites'] = $this->drupalRoot . '/core/tests/TestSuites';
+
+        $classMap = [
+            '\\Drupal\\Tests\\Core\\Render\\BubblingTest' => $this->drupalRoot . '/core/tests/Drupal/Tests/Core/Render/RendererBubblingTest.php',
+            '\\Drupal\\Tests\\Core\\Render\\PlaceholdersTest' => $this->drupalRoot . '/core/tests/Drupal/Tests/Core/Render/RendererTestBase.php',
+            '\\Drupal\\Tests\\Core\\Render\\TestAccessClass' => $this->drupalRoot . '/core/tests/Drupal/Tests/Core/Render/RendererTest.php',
+            '\\Drupal\\Tests\\Core\\Render\\TestCallables' => $this->drupalRoot . '/core/tests/Drupal/Tests/Core/Render/RendererTest.php',
+        ];
+        $this->autoloader->addClassMap($classMap);
     }
```
mglaman, line 232: This didn't fix the internal error:

```
Error: Internal error: Internal error: Method of class Drupal\Tests\Core\Render\BubblingTest cannot be used as the class does not exist in file /home/runner/work/_temp/drupal/core/tests/Drupal/Tests/Core/Render/RendererBubblingTest.php
```

I don't know what to do.

mglaman (changes requested):
We need to run PHPStan with phpstan-drupal and this PR against Drupal core using `--debug`, at least against `RendererPlaceholdersTest` to get the full stacktrace. I don't know why PHPStan can detect but PHP's internal reflection breaks.

PR #567, opened 2023-05-13
Replace `set-output` commands

### LINKED ISSUE #566: Replace `set-output` commands
# Feature request

GitHub is deprecating the `set-output` command.
>  Starting 1st June 2023 workflows using save-state or set-output commands via stdout will fail with an error.
See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This is used in build step "set the version alias for self".
Let's replace it with the recommended `GITHUB_OUTPUT` environment files:

Boegie:
Resolves #566

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -119,9 +119,9 @@ jobs:
       - name: "set the version alias for self"
         run: |
           if [ "${{ github.event_name }}" == 'pull_request' ]; then
-            echo ::set-output name=VERSION_ALIAS::dev-"${{ github.sha }}"
+            echo "VERSION_ALIAS=dev-"${{ github.sha }}"" >> $GITHUB_OUTPUT
           else
-            echo ::set-output name=VERSION_ALIAS::dev-main
+            echo "VERSION_ALIAS=dev-main" >> $GITHUB_OUTPUT
           fi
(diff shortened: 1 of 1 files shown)


Boegie:
Looking at the output of "require phpstan-drupal" I don't see the ` "${{ steps.branch_alias.outputs.VERSION_ALIAS }} as 1.1.99"` in it.
Neither before or after this PR.

Is this a case of "If it ain't broke don't fix it"?

mglaman (changes requested):
PR needs updating for `core_baseline` as well

mglaman:
It looks like the following:

```
COMPOSER_MEMORY_LIMIT=-1 composer require mglaman/phpstan-drupal "dev-main as 1.1.99" phpstan/extension-installer --with-all-dependencies
```

That was on a schedule, here's from the PR

```
COMPOSER_MEMORY_LIMIT=-1 composer require mglaman/phpstan-drupal "dev-05d6e8c5420c24468041d8fcdebf7f3860a47c13 as 1.1.99" phpstan/extension-installer --with-all-dependencies
```

Ensures sha is added after `dev-` for pulling in local merge

mglaman (approved):

PR #588, opened 2023-07-12
Add translatable interface stub for content entities

simonbaese:
This is a stub for the `TranslatableInterface`. Note that this stub will be unnecessary for Drupal 10.1+ (see https://www.drupal.org/project/drupal/issues/3269177). Nevertheless, it may be useful for users that run a prior version of Drupal.

This change is a little bit aggressive because it creates a stub for the `TranslatableInterface` directly, although only tests for content entities are provided. We could be more passive by basing the stub on the `ContentEntityBase`.

### DIFF
--- a/stubs/Drupal/Core/Entity/TranslatableInterface.stub
+++ b/stubs/Drupal/Core/Entity/TranslatableInterface.stub
@@ -1,0 +1,25 @@
+<?php
+
+namespace Drupal\Core\TypedData;
+
+interface TranslatableInterface {
+
+  /**
+   * @param string $langcode
+   * @return static
+   */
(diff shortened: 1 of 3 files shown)


simonbaese:
Need some advice on how to approach the failure in the baseline test.

mglaman (approved):
I think it's fine to provide the stub for `TranslatableInterface`. I think it's worth backporting the fix here for others.

As for the baseline failures for Drupal core: we can ignore those, I noticed that as well. There was a change in PHPStan I guess which broke the baseline. Or only in this project's CI.

mglaman:
Thank you!

PR #589, opened 2023-07-18
The entity property of an entity reference item can be null

larowlan:
If someone deletes the entity, we don't do cascade referential integrity in Drupal

We had this before in https://github.com/mglaman/phpstan-drupal/pull/356 but it was lost at some point


larowlan:
Hmm, ignore me there's a `?` there :joy:

PR #591, opened 2023-07-24
Add rule for unnecessary dependency injection in controllers

simonbaese:
**Motivation**
I often see unnecessary dependency injection in controllers that extend the controller base. With this new rule we can check whether a parent method call is sufficient. This idea could be extended to other use cases. For example when a class or a parent class uses a trait, e.g., `LoggerChannelTrait`, and still dependency injection is used for the corresponding service.

**Proposed changes**
Add new rule as proposed in pull request.

**Alternatives considered**
None so far.

**Testing steps**
Please see the attached test case.

### DIFF
--- a/src/Rules/Drupal/UnnecessaryDependencyInjectionRule.php
+++ b/src/Rules/Drupal/UnnecessaryDependencyInjectionRule.php
@@ -1,0 +1,98 @@
+<?php declare(strict_types = 1);
+
+namespace mglaman\PHPStanDrupal\Rules\Drupal;
+
+use Drupal\Core\Controller\ControllerBase;
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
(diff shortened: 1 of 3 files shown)


mglaman:
There is one problem: [`ControllerBase`](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Controller/ControllerBase.php) has these helper methods _without_ using dependency injection.

```php
  protected function entityTypeManager() {
    if (!isset($this->entityTypeManager)) {
      $this->entityTypeManager = $this->container()->get('entity_type.manager');
    }
    return $this->entityTypeManager;
  }
```

The container isn't injected, either.

```php
  private function container() {
    return \Drupal::getContainer();
  }
```

Generally, I am against anyone using `ControllerBase` and instead scaffolding their controllers and dependency injection for this reason. But, if using, they still should inject the services so the controller can be tested.

simonbaese:
Yes, the more I think about this the least I like the idea. Same goes for the traits.

Had some fun playing around. Let's close this.

PR #592, opened 2023-07-24
Check also for Node\Scalar\Encapsed

claudiu-cristea:
**Motivation**

In my project I'm getting:

> Missing cache backend declaration for performance.

with a plugin manager having the cache backend line in the constructor:

```php
$this->setCacheBackend($cache_backend, "mymodule_{$type}_plugins");
```

where `$type` is passed in constructor's signature, same as in Views plugin manager.

**Proposed changes**

This PR

**Alternatives considered**

None.

**Testing steps**

Unfortunately, for some mysterious reasons, I can't replicate this in the test from phpstan-drupal. I'm missing something. But applying the PR as patch to my project solves the issue.

I'm using

* PHPStan 1.10.25
* mglaman/phpstan-drupal 1.1.36

### DIFF
--- a/src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
+++ b/src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
@@ -59,7 +59,7 @@ class PluginManagerSetsCacheBackendRule extends AbstractPluginManagerRule
                     throw new ShouldNotHappenException();
                 }
                 $cacheKey = $setCacheBackendArgs[1]->value;
-                if (!$cacheKey instanceof Node\Scalar\String_ && !$cacheKey instanceof Node\Scalar\Encapsed) {
+                if (!$cacheKey instanceof Node\Scalar\String_) {
                     continue;
                 }
                 $hasCacheBackendSet = true;
--- a/src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
+++ b/src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
@@ -5,7 +5,6 @@ namespace mglaman\PHPStanDrupal\Rules\Drupal\PluginManager;
 use PhpParser\Node;
 use PhpParser\Node\Stmt\ClassMethod;
 use PHPStan\Analyser\Scope;
-use PHPStan\ShouldNotHappenException;
 use PHPStan\Type\Type;
 
 class PluginManagerSetsCacheBackendRule extends AbstractPluginManagerRule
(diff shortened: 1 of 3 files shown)


src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
```diff
@@ -59,7 +59,7 @@ public function processNode(Node $node, Scope $scope): array
                     throw new ShouldNotHappenException();
                 }
                 $cacheKey = $setCacheBackendArgs[1]->value;
-                if (!$cacheKey instanceof Node\Scalar\String_) {
+                if (!$cacheKey instanceof Node\Scalar\String_ && !$cacheKey instanceof Node\Scalar\Encapsed) {
                     continue;
                 }
                 $hasCacheBackendSet = true;
```
mglaman, line 65: This whole thing is bad :( It's old code. It needs to be updated for PHPStan's type system. Something like


```php
$type = $scope->getType($setCacheBackendArgs[1]->value);
foreach ($type->getConstantStrings() as ...)
```

mglaman (commented):
I'll try to look at it this week

mglaman:
Looks like this rule doesn't have a test, either 😬. I'll get that added.

claudiu-cristea:
Thank you. I've reverted the "fix" from [f693958](https://github.com/mglaman/phpstan-drupal/pull/592/commits/f693958055f8cf23fdec28a7ae8eccdded3b08db) and extended the test cases with the ones revealing the bug. Now those tests are failing as I've expected. We have the proof!

claudiu-cristea:
Not quite sure but I think Drupal core's exception here https://git.drupalcode.org/project/drupal/-/blob/11.x/core/phpstan.neon.dist#L37 was caused exactly by this issue as would make fail, at least, this plugin manager https://git.drupalcode.org/project/drupal/-/blob/11.x/core/modules/views/src/Plugin/ViewsPluginManager.php

longwave:
Would like to be able to remove that line from core's phpstan.neon, if I remember correctly there were way too many false positives and that is why it was originally added.

src/Rules/Drupal/PluginManager/PluginManagerSetsCacheBackendRule.php
```diff
@@ -53,29 +54,30 @@ public function processNode(Node $node, Scope $scope): array
...
 
+                $cacheKey = array_map(
+                    static fn (Type $type) => $type->getValue(),
+                    $scope->getType($setCacheBackendArgs[1]->value)->getConstantStrings()
+                );
+                if (count($cacheKey) === 0) {
+                    continue;
+                }
```
mglaman, line 69: This should fix all the false positives. The problem is that `$hasCacheBackendSet` wasn't set unless the `$cache_key` value could be resolved. When really that's only used for the cache tag inspection.

claudiu-cristea:
Thank you, it was so fast 👍

Issue #585, opened 2023-06-30
Deprecated scope resolver for `@group legacy`


mglaman:
# Feature request

This should go into phpstan-symfony, but we'll bring it here first.

```
class GroupLegacyScopeResolver implements DeprecatedScopeResolver
{

	public function isScopeDeprecated(Scope $scope): bool
	{
		$function = $scope->getFunction();
		return $function !== null
			&& $function->getDocComment() !== null
			&& strpos($function->getDocComment(), '@group legacy') !== false;
	}

}
```

```
services:
	-
		class: GroupLegacyScopeResolver
		tags:
			- phpstan.deprecations.deprecatedScopeResolver
```

PR #595, opened 2023-08-09
Add `@group legacy` deprecated scope resolver

mglaman:
Fixes https://github.com/mglaman/phpstan-drupal/issues/585

Adds `phpstan/phpstan-deprecation-rules` as a direct dependency and minimum version of 1.1.4.

This will cause the 1.2.0 release of phpstan-drupal.

Drupal core can use the following to prevent the deprecation rules extension from registering:

```json
{
  "extra": {
    "phpstan/extension-installer": {
      "ignore": [
        "phpstan/phpstan-deprecation-rules"
      ]
    }
  }
}
```


bbrala:
Why do we need a direct dependency, it is contained in phpstan?

mglaman:
```
use PHPStan\Rules\Deprecations\DeprecatedScopeResolver;
```

This interface is provided by the deprecations rules package. Not PHPStan itself.

PR #597, opened 2023-08-20
Detect public static $modules when it should be protected static $modules in tests

### LINKED ISSUE #463: Detect `public static $modules` when it should be `protected static $modules` in tests

Boegie:
Fixes #463

### DIFF
--- a/src/Rules/Drupal/TestClassesProtectedPropertyModulesRule.php
+++ b/src/Rules/Drupal/TestClassesProtectedPropertyModulesRule.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace mglaman\PHPStanDrupal\Rules\Drupal;
+
+use PhpParser\Node;
+use PHPStan\Analyser\Scope;
+use PHPStan\Node\ClassPropertyNode;
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+use PHPStan\ShouldNotHappenException;
+use PHPUnit\Framework\TestCase;
+
+class TestClassesProtectedPropertyModulesRule implements Rule
+{
+
+    public function getNodeType(): string
+    {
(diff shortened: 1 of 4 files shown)


src/Rules/Drupal/TestClassesProtectedPropertyModulesRule.php
```diff
@@ -0,0 +1,72 @@
...
+     * @return bool
+     *   True, if the given class reflection extends PPHPUnit's TestCase, false
+     *   otherwise.
+     */
+    protected function extendsPHPUnitFrameworkTestCase(ClassReflection $classReflection): bool
+    {
+        return
+            !$classReflection->isAnonymous()
+            && in_array(TestCase::class, $classReflection->getParentClassesNames(), true);
+    }
```
mglaman, line 71: instead of making this its own method, can we just have the conditions split out above?

```php
if ($classReflection->isAnonymous()) {
    return [];
}
if (!in_array(TestCase::class, $classReflection->getParentClassesNames(), true)) {
    return [];
}
```

src/Rules/Drupal/TestClassesProtectedPropertyModulesRule.php
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        if ($node->getName() !== 'modules') {
+            return [];
+        }
+
+        if ($node->isPublic()) {
+            return [
+                RuleErrorBuilder::message(
+                    sprintf('Declaring the ::$modules property as non-protected in %s is required.', $scopeClassReflection->getName())
```
mglaman, line 48: Should `$modules` be prefixed with `::`?

Can we just say:

```
The $modules property must be properted.
```

mglaman (changes requested):

mglaman (approved):

Boegie:
Thanks and sorry I never got the time to apply the requested changes myself, bit busy lately.

PR #606, opened 2023-09-27
Add information on why this package is needed

### LINKED ISSUE #601: Add "why is this needed" to README
# Feature request

It would be super if the README docs explained why PHPStan doesn't find Drupal symbols, and how this package fixes that omission.

mglaman:
Fixes #601


bkline:
Perfect!

mglaman:
Thanks!

Issue #608, opened 2023-10-01
Rule to check if test classes names end with "Test"


Boegie:
# Feature request

Over in https://www.drupal.org/project/drupal/issues/3268489 catch wrote:

> The ending with test feels like something we could add to phpstan - any non-abstract subclass of whatever the root test base class is. I think we should open a follow-up for that. These specific tests aren't going to be renamed again once this goes in and it'll stop regressions that would cause those tests to fail.

PR #609, opened 2023-10-01
TestClassSuffixNameRule

Boegie:
Trying to implement #608


Boegie:
Let's wait until somebody comments on https://www.drupal.org/project/drupal/issues/3268489#comment-15252653 about the  `abstract`-ness of `*TestBase` classes.

Boegie:
Verdict is in:

> The base classes should always be abstract yes, not sure whether it's documented but it's what should happen.
_catch in https://www.drupal.org/project/drupal/issues/3268489#comment-15252749_

For me the rule works as expected and is ready for review.

mglaman:
@Boegie I'll defer to your judgement if the baseline failures look OK. When you say it's good, I'll merge

Boegie:
Ohh, the pressure... ;)

Thanks @mglaman, I have something cooking locally that has three rules for test classes, one for the class name (this one), one for the namespace and one for the file location.

Since they both use a lot of the same code it makes sense to me to create a base class for this, thus trying to create one PR to rule them all. 

Shouldn't be long...
_#FamousLastWords_

PR #610, opened 2023-10-06
Add null to the container get method return type when ContainerInterace::NULL_ON_INVALID_REFERENCE is passed as the second argument.

mad-briller:
When developing modules that interact with optional modules it is useful to pass the `ContainerInterface::NULL_ON_INVALID_REFERENCE` second arg and do a null check rather than cumbersome exception handling 

This change adds `null` to the return type of any `$container->get` method calls that explicitly state it may return null

This also improves handling of `$container->get` with any second arg in general, as the second arg would disable the return type extension, causing it to fallback to the `object|null` return type on the ContainerInterface even though the module is likely available due to `require-dev` and this extension could provide a better type.

Looking at the [drupal source](https://github.com/drupal/drupal/blob/1325be71ae78c43818dd84dd2e542b21559435af/core/lib/Drupal/Component/DependencyInjection/Container.php#L131) it only respects 2 out of 5 of the constants provided by symfony but i added tests for all 5 to be sure

### DIFF
--- a/src/Type/ContainerDynamicReturnTypeExtension.php
+++ b/src/Type/ContainerDynamicReturnTypeExtension.php
@@ -9,9 +9,10 @@ use PHPStan\Reflection\MethodReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\Constant\ConstantBooleanType;
 use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\NullType;
 use PHPStan\Type\Type;
 use PHPStan\Type\TypeCombinator;
-use Psr\Container\ContainerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class ContainerDynamicReturnTypeExtension implements DynamicMethodReturnTypeExtension
 {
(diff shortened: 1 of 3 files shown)


mad-briller:
hmm i'm not sure what's going on with actions, the tests pass for me locally:

```
Ξ> ~/phpstan-drupal container-optional
=> php vendor/bin/phpunit
PHPUnit 9.6.13 by Sebastian Bergmann and contributors.

Warning:       No code coverage driver available

.........S..................................S..................  63 / 431 ( 14%)
..SS........................................................... 126 / 431 ( 29%)
............................................................... 189 / 431 ( 43%)
............................................................... 252 / 431 ( 58%)
............................................................... 315 / 431 ( 73%)
............................................................... 378 / 431 ( 87%)
.....................................................           431 / 431 (100%)

Time: 00:06.563, Memory: 366.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 431, Assertions: 773, Skipped: 4.
Ξ> ~/phpstan-drupal container-optional
=> php --version
PHP 7.4.33 (cli) (built: Sep  2 2023 08:03:46) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.33, Copyright (c), by Zend Technologies
    ```

File: src/Type/ContainerDynamicReturnTypeExtension.php [hunk 1]
```diff
@@ -41,24 +43,53 @@ public function getTypeFromMethodCall(
...
+
+            if (isset($args[1])) {
+                $invalidBehaviour = $scope->getType($args[1]->value);
+
+                if ($invalidBehaviour instanceof ConstantIntegerType
+                    && $invalidBehaviour->getValue() === ContainerInterface::NULL_ON_INVALID_REFERENCE
+                ) {
+                    $types[] = new NullType();
+                }
+            }
```
mglaman, line 80: To avoid an instance check, is there a matching getIntegerTypes or something to foreach over?

mglaman (commented):
This looks great, thanks!

src/Type/ContainerDynamicReturnTypeExtension.php
mad-briller: found the `getConstantScalarValues` method on the `Type` interface.

This means it can handle compound types aswell now so great question

mad-briller:
hey @mglaman is there anything left for me to do on this pr? thanks

mglaman:
I think it's all set, just need to sit down and merge during my open source time

mglaman (approved):

PR #615, opened 2023-10-29
Add decorator support to ServiceMap.

### LINKED ISSUE #614: The ServiceMap does not process decorators
# Bug report

Decorators are not currently processed for the service map.

Due to how the ServiceMap processes files there is no way to know what modules are enabled, therefor we can't attempt to determine what the highest priority service is to return that class, however we can return all possible options as a UnionType, which may actually be the better choice as it forces developers to understand that multiple services may be involved (preventing polymorphic calls).

In a Drupal KernelTest.
```
    $user_auth = $this->container->get('user.auth');
    $this->assertInstanceOf(TfaUserAuth::class, $user_auth);
```
Assuming that the user.auth service is decorated by TfaUserAuth::class, PHPStan incorrectly reports that assertInstanceOf will always be FALSE as the class is UserAuth.

cmlara:
Fixes #614

### DIFF
--- a/src/Drupal/DrupalServiceDefinition.php
+++ b/src/Drupal/DrupalServiceDefinition.php
@@ -50,7 +50,7 @@ class DrupalServiceDefinition
      */
     private $decorators = [];
 
-  public function __construct(string $id, ?string $class, bool $public = true, ?string $alias = null)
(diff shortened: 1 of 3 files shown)


mglaman (approved):
Thanks for this!

PR #620, opened 2023-11-08
Use TypeCombinator::union() to combine decorated service classes.

cmlara:
This should fix the issues reported in https://github.com/apigee/apigee-edge-drupal/pull/969


src/Drupal/DrupalServiceDefinition.php
```diff
@@ -122,7 +122,7 @@ public function getType(): Type
             foreach ($decorating_services as $service_id => $service_definition) {
                 $combined_services[] = $service_definition->getType();
             }
-            return new UnionType($combined_services);
+            return call_user_func_array([TypeCombinator::class, "union"], $combined_services);
```
mglaman, line 125: Does the following work instead of `call_user_func_array`

```php
TypeCombinator::union(...$combined_services)
```

mglaman (approved):

PR #622, opened 2023-11-08
Do not error when a service decorates an unknown service.

### LINKED ISSUE #618: Update to 1.2.1 undefined array key with PHPTuf etc.
# Bug report

Hi all, since update to 1.2.1 we have the following issue. With version 1.2.0 is working fine! The following output helps. The issue is in the new integration on ServiceMap.php

```
[user@server:~/www/domain.com/www.domain.com/webroot] $ TMPDIR=/tmp/.phpstan/drupalchecker_6549ec7bb9f7b2 ../vendor/bin/drupal-check -d modules/ -n --format=json --no-progress -vvv
Performing deprecation checks
Analyzing path: /home/user/www/domain.com/www.domain.com/webroot/modules
Current working directory: /home/user/www/domain.com/www.domain.com/webroot
Using Drupal root: /home/user/www/domain.com/www.domain.com/webroot
Using vendor root: /home/user/www/domain.com/www.domain.com/vendor
Assumed running as global dependency
PHPStan path: /home/user/www/domain.com/www.domain.com/vendor/phpstan/phpstan/phpstan.phar
PHPStan configuration path: /tmp/.phpstan/drupalchecker_6549ec7bb9f7b2/drupal_check_phpstan_1699344235.neon
PHPStan configuration:
parameters:
        tipsOfTheDay: false
        reportUnmatchedIgnoredErrors: false
        excludePaths:
                - */tests/Drupal/Tests/Listeners/Legacy/*
                - */tests/fixtures/*.php
                - */settings*.php
                - */node_modules/*

        ignoreErrors:
                - '#\Drupal calls should be avoided in classes, use dependency injection instead#'
                - '#Plugin definitions cannot be altered.#'
                - '#Missing cache backend declaration for performance.#'
                - '#Plugin manager has cache backend specified but does not declare cache tags.#'
                - '#Unsafe usage of new static\(\)#'

        drupal:
                drupal_root: /home/user/www/domain.com/www.domain.com/webroot

        level: 2
        bootstrapFiles:
                - /home/user/www/domain.com/www.domain.com/vendor/mglaman/drupal-check/error-bootstrap.php


Executing PHPStan
PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/user/www/domain.com/www.domain.com/webroot/modules/contrib/image_effects/image_effects.module on line 190
Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /home/user/www/domain.com/www.domain.com/webroot/modules/contrib/image_effects/image_effects.module on line 190
PHP Warning:  Undefined array key "PhpTuf\ComposerStager\API\Finder\Service\ExecutableFinderInterface" in /home/user/www/domain.com/www.domain.com/vendor/mglaman/phpstan-drupal/src/Drupal/ServiceMap.php on line 62
Warning: Undefined array key "PhpTuf\ComposerStager\API\Finder\Service\ExecutableFinderInterface" in /home/user/www/domain.com/www.domain.com/vendor/mglaman/phpstan-drupal/src/Drupal/ServiceMap.php on line 62
Error thrown in /home/user/www/domain.com/www.domain.com/vendor/mglaman/phpstan-drupal/src/Drupal/ServiceMap.php on line 62 while loading bootstrap file /home/user/www/domain.com/www.domain.com/vendor/mglaman/phpstan-drupal/drupal-autoloader.php: Call to a member function addDecorator() on null
Finished executing PHPStan
Unlinking PHPStan configuration
Return PHPStan exit code
Thanks for using drupal-check!

Consider sponsoring the development of the maintainers which make drupal-check possible:

- phpstan (ondrejmirtes): https://github.com/sponsors/ondrejmirtes
- phpstan-deprecation-rules (ondrejmirtes)): https://github.com/sponsors/ondrejmirtes
- phpstan-drupal (mglaman)): https://github.com/sponsors/mglaman
- drupal-check (mglaman)): https://github.com/sponsors/mglaman
```

cmlara:
Fixes #618

### DIFF
--- a/src/Drupal/ServiceMap.php
+++ b/src/Drupal/ServiceMap.php
@@ -59,6 +59,9 @@ class ServiceMap
 
         foreach ($decorators as $decorated_service_id => $services) {
             foreach ($services as $dcorating_service_id) {
+                if (!isset(self::$services[$decorated_service_id])) {
+                    continue;
+                }
                 self::$services[$decorated_service_id]->addDecorator(self::$services[$dcorating_service_id]);
             }
         }
--- a/tests/src/ServiceMapFactoryTest.php
+++ b/tests/src/ServiceMapFactoryTest.php
@@ -103,7 +103,11 @@ final class ServiceMapFactoryTest extends TestCase
             'service_map.deocrating_base' => [
                 'decorates' => 'service_map.base_to_be_decorated',
                 'class' => 'Drupal\service_map\SecondBase',
-            ]
+            ],
+            'decorating_an_unknown_service' => [
+                'decorates' => 'unknown',
+                'class' => 'Drupal\service_map\Override',
+            ],
         ]);
         $validator($service->getService($id));
     }


mglaman (commented):
Is this a bug? I mean it's a legitimate error if their services.yml is decorating an unknown service.. unless it's a service defined at runtime via a service provider (which this package doesn't handle)

cmlara:
I didn't see anything that made it apparent to me that the ServiceMap is intended to validate that services.yml's are well formed with all dependencies met.

Currently the ServiceMap ignores services with an unknown parent, while this change causes us to not attempt to enrich and unknown decorated service (we don't need to ignore the entire service, we just can't provide richer data to the decorated service.)

The ServiceProvider was a final tipping point since we already know nothing about services provided by a ServiceProvider  making this a hard fail would add a dependency that we be able to process ServiceProviders.

I also considered that when $container->get('unknown') is called the NULL from the ServiceMap doesn't appear to be reaching into PHPStan, this means the unknown service are currently treated as MIXED elsewhere.

If we don't want to silently ignore decorators we would likely want to refactor this to at least raise a clear exception instead of an undefined array key access.

mglaman (approved):

mglaman:
Let's silently ignore. I've been trying to find ways to make this "better." The "normal way" would be reading a dumped service container – like Symfony can do. But with Drupal that requires a database and more.

PR #634, opened 2023-11-11
Implement a rule to check deprecated hook implementations

### LINKED ISSUE #126: Deprecated hooks not flagged

Kingdutch:
This will analyse hook implementations in `.inc` and `.module` files. It detects a hook by looking at the pattern `mymodule_[hook]` which is detected from the file-name the function is in.

It will replace `mymodule` with `hook` and then try to find a function with that name. For this to work PHPStan Drupal's `DrupalAutoloader` class must be adjusted to load `.api.php` files as well so that the hooks can be discovered. In testing on the Open Social code-base this did not cause any issues.

The rule will provide the deprecation message in the error if it's provided, if no deprecation message was provided (i.e. a lonesome `@deprecated` tag) then the hook will simply be shown as deprecated without advice.

Fixes #126

### DIFF
--- a/src/Rules/Deprecations/DeprecatedHookImplementation.php
+++ b/src/Rules/Deprecations/DeprecatedHookImplementation.php
@@ -1,0 +1,68 @@
+<?php
+
+namespace mglaman\PHPStanDrupal\Rules\Deprecations;
+
+use PhpParser\Node;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Function_;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+class DeprecatedHookImplementation implements Rule
+{
+
+    /**
+     * @var \PHPStan\Reflection\ReflectionProvider
+     */
+    protected $reflectionProvider;
+
+    public function __construct(ReflectionProvider $reflectionProvider)
+    {
+        $this->reflectionProvider = $reflectionProvider;
+    }
+
+    public function getNodeType(): string
+    {
+        return Function_::class;
+    }
+
+    public function processNode(Node $node, Scope $scope) : array
+    {
+        assert($node instanceof Function_);
+        if (!str_ends_with($scope->getFile(), ".module") && !str_ends_with($scope->getFile(), ".inc")) {
+            return [];
+        }
+
+        // We want both name.module and name.views.inc, to resolve to name.
+        $module_name = explode(".", basename($scope->getFile()))[0];
+
+        // Hooks start with their own module's name.
+        if (!str_starts_with($node->name->toString(), "{$module_name}_")) {
+            return [];
+        }
+
+        $function_name = $node->name->toString();
+        $hook_name = substr_replace($function_name, "hook", 0, strlen($module_name));
+
+        $hook_name_node = new Name($hook_name);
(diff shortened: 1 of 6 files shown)


src/Rules/Deprecations/DeprecatedHookImplementation.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function getNodeType(): string
+    {
+        return Function_::class;
+    }
+
+    public function processNode(Node $node, Scope $scope) : array
+    {
+        assert($node instanceof Function_);
+        if (!str_ends_with($scope->getFile(), ".module") && !str_ends_with($scope->getFile(), ".inc")) {
```
mglaman, line 34: phpstan-drupal still supports PHP 7.4 technically for Drupal 9. But we can trust the PHP 8 polyfill exists for D9, so it should be okay right?

mglaman (commented):
Awesome! This looks great

mglaman:
I am surprised this works given

```
        excludePaths:
                - '*.api.php'
```

src/Rules/Deprecations/DeprecatedHookImplementation.php
bbrala: Also, Drupal 9 officially doesn't support 7.4 anymore ;)

mglaman: "officially"

mglaman:
@Kingdutch 

> For this to work PHPStan Drupal's DrupalAutoloader class must be adjusted to load .api.php files as well so that the hooks can be discovered. In testing on the Open Social code-base this did not cause any issues.

I don't see these changes? We'd have to add something here https://github.com/mglaman/phpstan-drupal/blob/main/src/Drupal/DrupalAutoloader.php#L110-L113 which is how we load post update files

src/Rules/Deprecations/DeprecatedHookImplementation.php
bbrala: Probably no breaking code, haven't noticed, just bo promised I think

Kingdutch: I've used `str_ends_with` and `str_starts_with` for pretty much the entirety of Drupal 9's lifetime :D The `symfony/polyfill-php80` package takes care of the functions (and Drupal depends on it).

Not sure if we need to add a dependency on that package explicitly in phpstan-drupal?

Kingdutch:
Modified the DrupalAutoloader for `.api.php` discovery (I suspect I had that modified locally but it went missing because it wasn't in the `open_social` repo where I originally committed this change).

Moved the hook declarations to the fixture's `.api.php` file to prove that works :)

src/Rules/Deprecations/DeprecatedHookImplementation.php
mglaman: Nah, I think we're OK and I should probably just drop PHP 7.4. I was waiting for PHPStan to drop it, but the Drupal community has pretty much moved onto PHP 8.1 (it seems.)

Nothing to do here, though.

mglaman (approved):
This is amazing! I'll merge later today.

PR #641, opened 2023-11-14
Make it possible to opt out of loading api.php files

### LINKED ISSUE #640: Including api.php in DrupalAutoloader throws PHP redeclare error
# Bug report

In the latest 1.2.3, changes were made to the `DrupalAutoloader.php` to include `.api.php` files: https://github.com/mglaman/phpstan-drupal/commit/446a99004db28b5d711d9e834f0818c427f5be40

This causes a PHP error we can't fix or ignore because it's between the core `field` module and the contrib `tmgmt_content`. Both modules include the same hook example in their API file so requiring those files will throw a PHP Fatal error.

```
PHP Fatal error:  Cannot redeclare hook_field_info_alter() (previously declared in /projects/site/web/core/modules/field/field.api.php:53) in /projects/site/web/modules/contrib/tmgmt/sources/content/tmgmt_content.api.php on line 47
Fatal error: Cannot redeclare hook_field_info_alter() (previously declared in /projects/site/web/core/modules/field/field.api.php:53) in /projects/site/web/modules/contrib/tmgmt/sources/content/tmgmt_content.api.php on line 47
```

eiriksm:
This makes it possible to opt out of loading api.php files, which can be something you want if you are experiencing the issue described in #640 

I tried to add a test for this as well, but since we have a test that requires us to scan for the file, adding it to a file in the fixture dir would make other tests fail :) I guess we could move some files in there as part of ::setup and then remove them in ::teardown or something. But feels hacky. Open to adding it, if adding test coverage would make this merged ;)

fixes #640


src/Drupal/DrupalAutoloader.php
```diff
@@ -112,7 +112,8 @@ public function register(Container $container): void
                 $this->loadAndCatchErrors($module_dir . '/' . $module_name . '.post_update.php');
             }
             // Add .api.php
-            if (file_exists($module_dir . '/' . $module_name . '.api.php')) {
+            $opt_out_of_api_php_scanning = isset($drupalParams['opt_out_api.php']) && (bool) $drupalParams['opt_out_api.php'] === true;
+            if (!$opt_out_of_api_php_scanning && file_exists($module_dir . '/' . $module_name . '.api.php')) {
                 $this->loadAndCatchErrors($module_dir . '/' . $module_name . '.api.php');
```
mglaman, line 117: Technically, it will always be set due to schema and defaults. Can `opt_out_of_api_php_scanning` be defined outside of the `foreach` loop.

extension.neon
```diff
@@ -233,6 +234,7 @@ parameters:
 parametersSchema:
 	drupal: structure([
 		drupal_root: string()
+		opt_out_api.php: boolean()
```
mglaman, line 237: I'd like to change this to:

```
bleedingEdge: structure[
	checkDeprecatedHooks: boolean()
])
```

Then the default is

```
drupal_root: '% currentWorkingDirectory%'
bleedingEdge:
	checkDeprecatedHooks: false
```

mglaman (commented):
Some thoughts. I can make the changes later in my afternoon. But noting them here in case you have bandwidth.

eiriksm:
Alrighty, I changed it up now. Still unsure how you would use the bleeding edge concept, so no logic for that is applied here

eiriksm:
Right, now the test added for the api.php functionality fails, but I am sure you can rework or remove that somehow? I will leave in ~30 mins, so feel free to take over :v:

mglaman:
99% tests include this config file https://github.com/mglaman/phpstan-drupal/blob/main/tests/fixtures/config/phpunit-drupal-phpstan.neon so it can be turned on there

I'll poke at this later, thanks for the changes!

extension.neon
```diff
@@ -233,6 +235,9 @@ parameters:
 parametersSchema:
 	drupal: structure([
 		drupal_root: string()
+		bleedingEdge: structure([
+			checkDeprecatedHooks: boolean()
+		])
```
mglaman, line 240: I want to rename this about "loading API files to check for deprecated hooks" so we can keep the rule and enable configured deprecated hooks manually

bleedingEdge.neon
```diff
@@ -0,0 +1,4 @@
+parameters:
+	drupal:
+		bleedingEdge:
+			checkDeprecatedHooks: true
```
mglaman, line 4: File needs to be documented

src/Rules/Deprecations/DeprecatedHookImplementation.php
```diff
@@ -46,6 +43,36 @@ public function processNode(Node $node, Scope $scope) : array
...
+                )->build()
+            ];
+        }
+        if (str_starts_with($hook_name, 'hook_field_widget_multivalue_') && str_ends_with($hook_name, '_form_alter')) {
+            return [
+                RuleErrorBuilder::message(
+                    "Function $function_name implements $hook_name which is deprecated in drupal:9.2.0 and is removed from drupal:10.0.0. Use hook_field_widget_complete_WIDGET_TYPE_form_alter instead.",
+                )->build()
+            ];
+        }
```
mglaman, line 74: This needs a refactor. Ideally, it would be a better format as a static array in a helper method.

bleedingEdge.neon
mglaman: It can be done later: https://github.com/mglaman/phpstan-drupal/issues/644

tests/src/Rules/DeprecatedHookImplementationTest.php
```diff
@@ -35,6 +35,18 @@ public function testRule() : void {
...
+            [
+                'Function deprecated_field_widget_hooks_field_widget_form_alter implements hook_field_widget_form_alter which is deprecated in drupal:9.2.0 and is removed from drupal:10.0.0. Use
+  hook_field_widget_single_element_form_alter instead.',
+                5
+            ],
+            [
+                'Function deprecated_field_widget_hooks_field_widget_textfield_form_alter implements hook_field_widget_WIDGET_TYPE_form_alter which is deprecated in drupal:9.2.0 and is removed from drupal:10.0.0. Use hook_field_widget_single_element_WIDGET_TYPE_form_alter instead.',
+                9
+            ]
+        ]);
```
mglaman, line 49: This should only run on `> 9 && < 10`

mglaman (approved):

PR #659, opened 2023-11-23
Adds DeprecatedHelperVisitor

bbrala:
Adds DeprecatedHelperVisitor which removes PHPStan parsing of everything in the 3rd argument of the DeprecationHelper, making it ignore deprecation messages in there.


src/Parser/DeprecatedHelperVisitor.php
```diff
@@ -0,0 +1,40 @@
...
+class DeprecatedHelperVisitor extends NodeVisitorAbstract implements NodeVisitor
+{
+
+    private ?Node\Arg $deprecatedCall = null;
+
+    public function enterNode(Node $node)
+    {
+
+        if ($node instanceof Node\Expr\StaticCall) {
+            if ($node->class instanceof Node\Name\FullyQualified && $node->class->isFullyQualified() && $node->class->toString() === 'Drupal\Component\Utility\DeprecationHelper') {
```
mglaman: It's always an instance of FullyQualified even if imported, right? Pretty sure but double checking.

src/Parser/DeprecatedHelperVisitor.php
```diff
@@ -0,0 +1,39 @@
...
+            }
+        }
+
+        if ($this->deprecatedCall !== null && $node instanceof Node\Arg && $this->isSavedArgument($node)) {
+            $this->deprecatedCall = null;
+            return NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
+        }
+
+        return null;
+    }
```
mglaman, line 30: Is there a chance our property won't get unset for some reason?

src/Parser/DeprecatedHelperVisitor.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        if ($node instanceof Node\Expr\StaticCall) {
+            if ($node->class instanceof Node\Name\FullyQualified && $node->class->toString() === 'Drupal\Component\Utility\DeprecationHelper') {
+                $this->deprecatedCall = $node->getArgs()[2];
+                return null;
+            }
+        }
+
+        if ($this->deprecatedCall !== null && $node instanceof Node\Arg && $this->isSavedArgument($node)) {
+            $this->deprecatedCall = null;
```
mglaman, line 25: The null check is technically duplicated in the `isSavedArgument` method

src/Parser/DeprecatedHelperVisitor.php
```diff
@@ -0,0 +1,39 @@
...
+            return NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
+        }
+
+        return null;
+    }
+
+    private function isSavedArgument(Node\Arg $node): bool
+    {
+        if ($this->deprecatedCall !== null && $node->getAttributes() === $this->deprecatedCall->getAttributes()) {
+            return true;
```
mglaman, line 35: Okay so when we detect a static call on the methods, we fetch and stash the callable for the deprecated case. Then when then node parser reaches the argument we check and unset.

tests/src/Parser/DeprecatedHelperVisitor/data/deprecation-helper-test.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace GroupLegacy;
```
bbrala, line 3: This should be a more sane namespace i guess

src/Parser/DeprecatedHelperVisitor.php
```diff
@@ -0,0 +1,39 @@
...
+            if ($node->class instanceof Node\Name\FullyQualified && $node->class->toString() === 'Drupal\Component\Utility\DeprecationHelper') {
+                $this->deprecatedCall = $node->getArgs()[2];
+                return null;
+            }
+        }
+
+        if ($this->deprecatedCall !== null && $node instanceof Node\Arg && $this->isSavedArgument($node)) {
+            $this->deprecatedCall = null;
+            return NodeTraverser::DONT_TRAVERSE_CURRENT_AND_CHILDREN;
+        }
```
mglaman, line 27: This is the magic, yeah? PHPStan doesn't analyze anything in the tree? Seems problematic and instead we should try adding information if possible. Probably not, though.

src/Parser/DeprecatedHelperVisitor.php
bbrala: yeah i removed this one after your review

src/Parser/DeprecatedHelperVisitor.php
bbrala: Calling deprecation helper without arguments? But that is impossible. 

Perhaps if while moving into the function something changes the node, so the argument is no longer on the same line and therefor not equal? Not sure how, and if that is possible. 

But the effect would basically be if would check the array a few times extra, which should be quite cheap.

There might be a way to add some more checks when in the argument. Or do something like keeping a stack of the last static call and check if that call is nothing weird. 

```
        if ($node instanceof Node\Expr\StaticCall) {
            $this->staticCall = $node; // this and check this?
            if ($node->class instanceof Node\Name\FullyQualified && $node->class->toString() === 'Drupal\Component\Utility\DeprecationHelper') {
                $this->deprecatedCall = $node->getArgs()[2];
                return null;
            }
        }
```

But that seems like overkill.

src/Parser/DeprecatedHelperVisitor.php
bbrala: I was making phpstan happy, the property can be null.

src/Parser/DeprecatedHelperVisitor.php
bbrala: yes!

src/Parser/DeprecatedHelperVisitor.php
bbrala: Yeah, it stops parsing as discussed on slack.

tests/src/Parser/DeprecatedHelperVisitor/data/deprecation-helper-test.php
mglaman: It's good enough for the test data :)

bbrala:
Hi @mglaman , any chance you can revisit this? I understand the drawbacks, but we need some solution to skip code inside the deprecations helper before we start posting patches.

mglaman:
:/ sorry it's been in my TODOs



I'll get to this shortly. I'm thinking we can enable it behind a feature flag. The project update bot can opt in. I'm still concerned that it just halts analysis.

bbrala:
This has been superceeded by #714, yay for phpstan developer @ondrejmirtes (and you @mglaman ;))

PR #702, opened 2024-01-05
Allow to provide drupal_root via an environment variable.

AlexSkrypnyk:
This helps in cases when PHPStan is called outside of the main website location.


AlexSkrypnyk:
Integration tests are currently failing on `main`, so not related to this change.

src/Drupal/DrupalAutoloader.php
```diff
@@ -85,7 +85,7 @@ public function register(Container $container): void
          * @var array{drupal_root: string, bleedingEdge: array{checkDeprecatedHooksInApiFiles: bool}} $drupalParams
          */
         $drupalParams = $container->getParameter('drupal');
-        $drupalRoot = realpath($drupalParams['drupal_root']);
+        $drupalRoot = realpath(getenv('DRUPAL_ROOT') !== false ? getenv('DRUPAL_ROOT') : $drupalParams['drupal_root']);
         $finder = new DrupalFinder();
```
mglaman, line 89: So this would add a new environment variable called `DRUPAL_ROOT` that can be defined? I wonder, since [`bootstrap.inc`](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/includes/bootstrap.inc?ref_type=heads) is included via Composer's autoloading if we should use the `DRUPAL_ROOT` constant.

But I don't know if I want to rely on that global constant. It's preferred to use `app.root` container parameter (which we cannot use.)

PR #716, opened 2024-01-16
Add missing return in ClassExtendsInternalClassRule.

mad-briller:
Whilst having a poke around in the code for the sake of curiosity, i noticed this expression that is missing a return and as such does nothing.

If i look at the commit that introduced the issue, it appears as though the `->tip` calls were added in the same commit, so i'm not sure if the `return` is intended to be there, or if the entire conditional is now redundant and not necessary. Interestingly the tests don't change with or without the `return` so it's hard to say whether or not it should be there?

Perhaps given that it's been like this for ~3 years the correct fix is to remove the conditional altogether? If that is preferred let me know and i can update the pull request.


mglaman:
Oh it definitely is intended. Given most people are running and analyzing only Drupal code which is not in a generic namespace, this probably isn't needed but was done as a safeguard.

Great catch

mglaman (approved):

mglaman:
Thanks!

PR #717, opened 2024-01-17
Simplify access to reflections in SymfonyCmfRoutingInClassMethodSignatureRule and TestClassesProtectedPropertyModulesRule.

mad-briller:
PHPStan added the ability to access reflection objects from many of it's virtual nodes which guarantees they will be there, avoiding the need to do these checks.


mglaman (approved):

mglaman:
Thanks for helping clean up!

PR #718, opened 2024-01-18
Fully qualify all global function calls.

mad-briller:
Modern php has several optimizations it can apply to global functions at compile-time, but only when they are fully-qualified.

For example it can [translate](https://php.watch/articles/php-zend-engine-special-inlined-functions) several functions into specific opcodes which are faster than their function equivalent.

It can also do [compile-time evaluation](https://github.com/php/php-src/blob/43b62b7007adcf194abe549ccb64a6977bfd5c4e/ext/standard/basic_functions.stub.php#L1596-L1597) of constant operands for several functions.

Additionally, fully-qualifying global functions avoids a local namespace lookup and as such makes calls faster in general. The improvement will be pretty minor but every little helps, especially in situations where these functions may be called thousands of times in a single analysis.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -21,6 +21,7 @@
 	<rule ref="SlevomatCodingStandard.Namespaces.UnusedUses"/>
 	<rule ref="SlevomatCodingStandard.Namespaces.UselessAlias"/>
 	<rule ref="SlevomatCodingStandard.Namespaces.UseSpacing"/>
+	<rule ref="SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly"/>
 	<rule ref="SlevomatCodingStandard.PHP.UselessSemicolon"/>
 	<rule ref="SlevomatCodingStandard.PHP.UselessParentheses"/>
 </ruleset>
--- a/src/DeprecatedScope/DeprecationHelperScope.php
+++ b/src/DeprecatedScope/DeprecationHelperScope.php
@@ -8,6 +8,8 @@ use Drupal\Component\Utility\DeprecationHelper;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\MethodReflection;
 use PHPStan\Rules\Deprecations\DeprecatedScopeResolver;
+use function class_exists;
+use function count;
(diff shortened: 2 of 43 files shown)


mglaman:
Did you manually do this? I wonder if we could use Rector to automate this and if we should add a rule to make sure functions are always imported.

I know they are in PHPStan, it'd be worth checking how that project enforces this

mad-briller:
Yeah i did do it manually

not sure about the rector aspect with regards to the automation, as that would require either pre-commit hooks or for github actions to add commits onto opened prs to apply changes.

It would be easy to add a phpcs sniff to enforce they're always fully-qualified though, phpstan itself uses https://github.com/slevomat/coding-standard and the sniff is `SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFallbackGlobalName`

Adopting the full `slevomant/coding-standard` ruleset would likely be quite a large piece of work as the slevomat standard is fairly opinionated, but i could require-dev it and add the single sniff if you'd like.

mglaman:
Drupal core also uses parts of slevomat/coding-standard (https://git.drupalcode.org/project/coder/-/blob/8.3.x/composer.json?ref_type=heads#L21)

It's used by Drupal core https://git.drupalcode.org/project/drupal/-/blob/11.x/core/phpcs.xml.dist?ref_type=heads#L255

I think it'd be great to add it as a dev dependency and include that one rule to start

mad-briller:
Turns out `slevomat/coding-standards` was already installed to the repository :sweat_smile:, but ooking into it the rule does more than just enforce the use of `use function`:

https://github.com/slevomat/coding-standard/blob/master/doc/namespaces.md#slevomatcodingstandardnamespacesreferenceusednamesonly-

it also prevents the use of inline namespaces and a few others, which increases the size of the changeset. I've resolved all the violations it came up with in a separate commit to make it easier for you to review.

mglaman (approved):
Thanks for this, and the extra fixes. There wasn't really a pattern for using FQCN or not. This is a nice addition to standardize and clean things up.

PR #719, opened 2024-01-19
Add implements generics to all Rule implementations and remove associated assert() calls.

mad-briller:
This change allows phpstan to statically verify that the rules work correctly without having to get `assert()` involved, as `assert()` is a finicky beast whose implications can change based on the php configuration of the system it is run on.

### DIFF
--- a/src/Rules/Classes/ClassExtendsInternalClassRule.php
+++ b/src/Rules/Classes/ClassExtendsInternalClassRule.php
@@ -11,6 +11,9 @@ use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use function sprintf;
 
+/**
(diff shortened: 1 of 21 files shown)


mglaman:
I'll check this out locally. I know I added a lot of the asserts for my own benefit with PhpStorm and less for PHPStan. But some of these were done before the current type system in PHPStan as well and may not be needed.

mad-briller:
Ah no worries, i'm using intelephense and that handles it well.

I know a few of the devs at my org are using phpstorm and they've said it's support for generics has improved a lot recently so it may support it now, let me know and i can close this PR if it's still an issue.

mglaman (approved):
Works like a charm, I appreciate the clean up!

PR #727, opened 2024-02-14
Make ^10.0 the default constraint for Drupal core

### LINKED ISSUE #711: Make ^10.0 the default constraint for Drupal core, update from ^9.0
Drupal 9 is going away. It can remain in the test run, but should be considered the edge case and no longer the default.

mglaman:
Fixes #711


src/Rules/Deprecations/SymfonyCmfRouteObjectInterfaceConstantsRule.php
```diff
@@ -26,6 +26,9 @@ public function getNodeType(): string
 
     public function processNode(Node $node, Scope $scope): array
     {
+        if (!class_exists(SymfonyRouteObjectInterface::class)) {
+            return [];
+        }
```
mglaman, line 31: This is causing `SymfonyCmfRouteObjectInterfaceConstantsRuleTest` to fail on ^9.0 and I'm not sure why.

PR #731, opened 2024-02-26
Add DependencySerializationTraitPropertyRule

### LINKED ISSUE #730: Form should not have private properties
# Feature request

When working with forms, you always should remember that the form can be in a multistep form or loaded by a AJAX window.
Private properties will not be serialized by the Drupal serializer due to complications with duplicate properties names.
So when your form gets serialized by the next step or AJAX, only protected, or higher available properties are set.
This leads to a lot of bugs and side effects, since unset properties hard error the page.

In most issues you can use encapsulation for fixing the unset variables, but in the first place and as a general rule you should always use protected or higher variables.
My proposal is that we don't allow private properties in FormInterface implementing classes to prevent the odd issues when going to AJAX or multistep.
But I am open for other suggestions.

MakerTim:
Adds a check for all FormInterface implementing classes that private variables are no longer allowed
See #730 for context and discussion 

Fixes #730

### DIFF
--- a/src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
+++ b/src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Rules\Drupal\Coder;
+
+use Drupal\Core\Form\FormInterface;
+use PhpParser\Node;
+use PHPStan\Node\ClassPropertyNode;
+use PHPStan\Analyser\Scope;
+use PHPStan\Rules\Rule;
+use PHPStan\Rules\RuleErrorBuilder;
+
+final class FormNoPrivatePropertiesRule implements Rule
+{
+
+    public function getNodeType(): string
+    {
+        return ClassPropertyNode::class;
+    }
+
+    public function processNode(Node $node, Scope $scope): array
+    {
+        if (!$node instanceof ClassPropertyNode
+            || !$node->getClassReflection()->implementsInterface(FormInterface::class)) {
+            return [];
+        }
+
+        if ($node->isPrivate()) {
+            return [
+                RuleErrorBuilder::message(
+                    'Private properties are note allowed on FormInterface classes due to seralisation.'
+                )->tip('See https://www.drupal.org/node/3110266')->build(),
+            ];
+        }
+        if ($node->isReadOnly()) {
+            return [
+                RuleErrorBuilder::message(
+                    'Private properties are note allowed on FormInterface classes due to seralisation.'
+                )->tip('See https://www.drupal.org/node/3110266')->build(),
+            ];
+        }
+
+        return [];
+    }
+
+}
--- a/src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
+++ b/src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 
 namespace mglaman\PHPStanDrupal\Rules\Drupal\Coder;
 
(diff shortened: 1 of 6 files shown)


src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
```diff
@@ -0,0 +1,44 @@
...
+                )->tip('See https://www.drupal.org/node/3110266')->build(),
+            ];
+        }
+        if ($node->isReadOnly()) {
+            return [
+                RuleErrorBuilder::message(
+                    'Private properties are note allowed on FormInterface classes due to seralisation.'
+                )->tip('See https://www.drupal.org/node/3110266')->build(),
+            ];
+        }
```
mglaman, line 40: Is this true? I thought read-only was only violated after being set. Or is it technically because it is set before __wakeup is called? Also, seems like a great candidate for its own rule class.

mglaman (commented):
Can we register the rule using a boolean in configuration like was done here: https://github.com/mglaman/phpstan-drupal/commit/20d27158b90bbdeeae9f299b335c057b0c200416

It's part of how I want to handle backwards compatibility when adding new rules.

src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
MakerTim: Yes, it has to do the way Drupal serializes the class.
`Cannot initialize readonly property CLASS::readonlyproperty from scope Drupal\Core\Form\FormBase`
from `/core/lib/Drupal/Core/DependencyInjection/DependencySerializationTrait.php:89`
image

And since it is triggering in both cases an invalid form, I suggest it shoud be on the same check.
check for valid form-classes for the serializer / serializable objects as a rule.

MakerTim:
I found the Drupal issue about the serialization issue.
https://www.drupal.org/project/drupal/issues/3110266#comment-15388060

But seeing the time it takes to fix this on the Drupal side of things, is this still the route the code should have been, specially if the serialization is interchangeable what the current direction is, then it should be correct in the first place.

src/Rules/Drupal/Coder/FormNoPrivatePropertiesRule.php
mglaman: That's fair! Can we have the rule return both errors then if they both exist instead of failing first on private and possibly failing again on readonly

PHPStan allows rules to return multiple errors

mglaman:
This actually effects anything using `DependencySerializationTrait`, so it can be made more generic.

src/Rules/Drupal/DependencySerializationTraitPropertyRule.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        if (!$node->getClassReflection()->hasTraitUse(DependencySerializationTrait::class)) {
+            return [];
+        }
+
+        $errors = [];
+        if ($node->isPrivate()) {
+            $errors[] = RuleErrorBuilder::message(
+                sprintf(
+                    'Private properties cannot be serialized with %s.',
```
mglaman, line 35: "cannot be serialized by"?

src/Rules/Drupal/DependencySerializationTraitPropertyRule.php
```diff
@@ -0,0 +1,50 @@
...
+                    DependencySerializationTrait::class
+                )
+            )->tip('See https://www.drupal.org/node/3110266')->build();
+        }
+        if ($node->isReadOnly()) {
+            $errors[] = RuleErrorBuilder::message(
+                sprintf(
+                    'Read only properties cannot be serialized with %s.',
```
mglaman, line 43: "Read only" or "Read-only" and "cannot be serialized by"?

src/Rules/Drupal/DependencySerializationTraitPropertyRule.php
mglaman: > Drupal\Core\DependencyInjection\DependencySerializationTrait does not support private properties.

src/Rules/Drupal/DependencySerializationTraitPropertyRule.php
mglaman: > Read-only properties are incompatible with Drupal\Core\DependencyInjection\DependencySerializationTrait

MakerTim:
Tested on local projects, great addition.

_Got stuck on writing tests and never completed a new commit_

mglaman:
Thanks for getting this started @MakerTim !

PR #737, opened 2024-03-20
AccessResultConditionTypeSpecifyingExtension

### LINKED ISSUE #734: Have AccessResult forbiddenIf and allowedIf perform truthy checks
# Feature request

PHPStan does checks on `if` statements to verify they're evaluating `true` or `false` and if the condition is always true or false.

It'd be great to recreate that rule for `\Drupal\Core\Access\AccessResult::allowedIf` and `\Drupal\Core\Access\AccessResult::forbiddenIf`

mglaman:
fixes #734

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -312,8 +312,11 @@ services:
 		class: mglaman\PHPStanDrupal\DeprecatedScope\GroupLegacyScope
 		tags:
 			- phpstan.deprecations.deprecatedScopeResolver
-
 	-
 		class: mglaman\PHPStanDrupal\DeprecatedScope\DeprecationHelperScope
 		tags:
 			- phpstan.deprecations.deprecatedScopeResolver
+	-
+		class: mglaman\PHPStanDrupal\Type\AccessResultConditionTypeSpecifyingExtension
+		tags:
+			- phpstan.typeSpecifier.staticMethodTypeSpecifyingExtension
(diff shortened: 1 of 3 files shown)


src/Rules/Drupal/AccessResultConditionRule.php
```diff
@@ -0,0 +1,81 @@
...
+                RuleErrorBuilder::message(sprintf(
+                    'Strict comparison using %s between %s and %s will always evaluate to false.',
+                    $condition->getOperatorSigil(),
+                    $leftType->describe(VerbosityLevel::value()),
+                    $rightType->describe(VerbosityLevel::value()),
+                ))->identifier(sprintf('%s.alwaysFalse', $condition instanceof Node\Expr\BinaryOp\Identical ? 'identical' : 'notIdentical'))->build(),
+            ];
+        }
+        return [];
+    }
```
mglaman, line 80: See `reportAlwaysTrueInLastCondition` in https://github.com/phpstan/phpstan-src/blob/f1cf3bb34d949f01266c9eb08097144b262f4e4a/src/Rules/Comparison/StrictComparisonOfDifferentTypesRule.php#L69-L97

except we probably always want to evaluate the last condition, because there isn't a chain. `reportAlwaysTrueInLastCondition` is false by default. So I'm not sure how to handle "if always true". I think.

PR #739, opened 2024-03-22
Add EntityRepositoryInterface return type extension

### LINKED ISSUE #514: Dynamic return type for EntityRepositoryInterface::loadEntityByUuid
Lets add an extension to dynamically detect the return type for `\Drupal\Core\Entity\EntityRepositoryInterface::loadEntityByUuid`

We already have `entityMapping` in phpstan.neon, I think the 'class' property from these can be safely reused for this purpose.

Presently doc is:

```
   * @return \Drupal\Core\Entity\EntityInterface|null
   *   The entity object, or NULL if there is no entity with the given UUID.
   */
  public function loadEntityByUuid($entity_type_id, $uuid);
```

We should be able to map the `$entity_type_id` param to an item from `entityMapping`, and dynamically return the class defined in `class` property or NULL.

e.g

```yaml
	drupal:
		entityMapping:
			node:
				class: Drupal\mymodule\Entity\Node\MyAbstractCustomNodeBase
				storage: Drupal\node\NodeStorage
```

effectively, if $entity_type_id === 'node':

```
public function loadEntityByUuid($entity_type_id, $uuid): ?\Drupal\mymodule\Entity\Node\MyAbstractCustomNodeBase;
```

As a bonus feature, maybe theres a way to statically detect if entity type supports uuid, and add the mapping. Since an exception is thrown when the entity type ID does not support UUID's.

mglaman:
fixes #514

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -273,6 +273,9 @@ services:
 		class: mglaman\PHPStanDrupal\Type\EntityStorage\EntityStorageDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 	-
+		class: mglaman\PHPStanDrupal\Type\EntityRepositoryReturnTypeExtension
+		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+	-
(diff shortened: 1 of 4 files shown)


dpi:
Thanks!

Issue #750, opened 2024-04-29
New rule for checking "static create()" method in controller


joshirohit100:
Controllers in Drupal-10.2.0 are auto-wired. Ref - https://www.drupal.org/node/3395716

To have a new rule to check this (its mostly DX)

(I know more general issue for different components - https://www.drupal.org/project/drupal/issues/3409787)

PR #751, opened 2024-04-29
Adding new rule for checking controller autowire

joshirohit100:
Issue - #750

### DIFF
--- a/src/Rules/Drupal/ControllerAutoWireRule.php
+++ b/src/Rules/Drupal/ControllerAutoWireRule.php
@@ -1,0 +1,72 @@
+<?php
+
+declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Rules\Drupal;
+
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node;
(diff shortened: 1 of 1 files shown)


src/Rules/Drupal/ControllerAutoWireRule.php
```diff
@@ -0,0 +1,72 @@
...
+            return [];
+        }
+
+        $namespace = $class->namespacedName->toCodeString();
+        $reflection = $this->reflectionProvider->getClass($namespace);
+        $extends_controllerbase = $reflection->isSubclassOf('Drupal\Core\Controller\ControllerBase');
+
+        if (!$extends_controllerbase) {
+            return [];
+        }
```
mglaman, line 64: This should check if the auto wire trait is used versus the specific base class

mglaman (changes requested):

src/Rules/Drupal/ControllerAutoWireRule.php
joshirohit100: Updated to check for the trait

PR #774, opened 2024-06-19
Seperate core and contrib api files

### LINKED ISSUE #770: Make opting into loading API files more granular to support finding deprecated hooks from Drupal core more easily
# Feature request

Follows up on #641 

Drupal Slack link: https://drupal.slack.com/archives/C03L6441E1W/p1718043125262999

Summary:

> :four: Detecting deprecated hooks with upgrade status module for Drupal 11 readiness

Added with #634. But then we had issues due to broken `.api.php` files somehow, made opt-in with #641.

Proposal: make current flag more granular so we load Drupal core `.api.php` files which shouldn't be broken.

1. Fix current docs, wording is wrong
2. Deprecate the current flag, treat it as controlling loading of contrib `.api.php` files, still defaults to `false`
3. Introduce a new flag for controlling loading of Drupal core `.api.php` files, defaults to `false`

Upgrade Status can opt-into loading core `.api.php` files.

bbrala:
Fixes #770

1. Added new options to bleeding edge. Current behaviour is the same.
2. Trigger deprecation notice when using old parameter, did not add a version since i dont know when it will be removed
3. Added defaults as per related issue. Behaviour is the same, but can be specified.
4. Wanted to add a test for a core module, but not sure how so i didnt.

### DIFF
--- a/bleedingEdge.neon
+++ b/bleedingEdge.neon
@@ -1,7 +1,7 @@
 parameters:
 	drupal:
 		bleedingEdge:
-			checkDeprecatedHooksInApiFiles: true
+			checkDeprecatedHooksInApiFiles: false
 			checkCoreDeprecatedHooksInApiFiles: true
 			checkContribDeprecatedHooksInApiFiles: true
 		rules:
--- a/bleedingEdge.neon
+++ b/bleedingEdge.neon
@@ -2,7 +2,7 @@ parameters:
 	drupal:
 		bleedingEdge:
 			checkDeprecatedHooksInApiFiles: true
-			checkCoreDeprecatedHooksInApiFiles: false
+			checkCoreDeprecatedHooksInApiFiles: true
 			checkContribDeprecatedHooksInApiFiles: false
 		rules:
 			testClassSuffixNameRule: true
--- a/bleedingEdge.neon
+++ b/bleedingEdge.neon
@@ -2,7 +2,7 @@ parameters:
 	drupal:
 		bleedingEdge:
 			checkDeprecatedHooksInApiFiles: true
-			checkCoreDeprecatedHooksInApiFiles: true
+			checkCoreDeprecatedHooksInApiFiles: false
 			checkContribDeprecatedHooksInApiFiles: false
 		rules:
 			testClassSuffixNameRule: true
--- a/bleedingEdge.neon
+++ b/bleedingEdge.neon
@@ -2,6 +2,8 @@ parameters:
 	drupal:
 		bleedingEdge:
 			checkDeprecatedHooksInApiFiles: true
+			checkCoreDeprecatedHooksInApiFiles: false
+			checkContribDeprecatedHooksInApiFiles: false
 		rules:
 			testClassSuffixNameRule: true
 			dependencySerializationTraitPropertyRule: true
--- a/bleedingEdge.neon
+++ b/bleedingEdge.neon
@@ -2,8 +2,8 @@ parameters:
 	drupal:
 		bleedingEdge:
 			checkDeprecatedHooksInApiFiles: true
-			checkCoreDeprecatedHooksInApiFiles: false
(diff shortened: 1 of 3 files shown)


Kingdutch (commented):
This looks good to me (I'm still a proponent of immediately setting `checkCoreDeprecatedHooksInApiFiles` to `true` but I know Matt wants to do it in a separate minor release).

Commenting rather than explicitly approving because I don't know if we can already remove `checkCoreDeprecatedHooksInApiFiles` anywhere. For example from `bleedingEdge.neon`, in order to prevent people from getting deprecation notices for something they load from the package itself.

Not entirely sure what the difference in behaviour between `bleedingEdge.neon` and `extension.neon` is though (since the values also differ in those two files).

bleedingEdge.neon
```diff
@@ -2,6 +2,8 @@ parameters:
 	drupal:
 		bleedingEdge:
 			checkDeprecatedHooksInApiFiles: true
+			checkCoreDeprecatedHooksInApiFiles: false
+			checkContribDeprecatedHooksInApiFiles: false
```
mglaman, line 6: In bleedingEdge these should be `true`, as it turns on all feature flags

extension.neon
```diff
@@ -242,6 +244,8 @@ parametersSchema:
 		drupal_root: string()
 		bleedingEdge: structure([
 			checkDeprecatedHooksInApiFiles: boolean()
```
mglaman, line 246: I wonder if there is some way to deprecated a flag

src/Drupal/DrupalAutoloader.php
```diff
@@ -127,7 +127,14 @@ public function register(Container $container): void
         $this->addThemeNamespaces();
         $this->registerPs4Namespaces($this->namespaces);
         $this->loadLegacyIncludes();
-        $checkDeprecatedHooksInApiFiles =  $drupalParams['bleedingEdge']['checkDeprecatedHooksInApiFiles'];
+
+        // Trigger deprecation error if checkDeprecatedHooksInApiFiles is enabled.
+        if ($drupalParams['bleedingEdge']['checkDeprecatedHooksInApiFiles']) {
+            trigger_error('The bleedingEdge.checkDeprecatedHooksInApiFiles parameter is deprecated and will be removed in a future release.', E_USER_DEPRECATED);
+        }
```
mglaman, line 134: ah this is how

bleedingEdge.neon
bbrala: Ok

src/Drupal/DrupalAutoloader.php
bbrala: Just missing version, but that's OK I think?

bbrala:
> This looks good to me (I'm still a proponent of immediately setting `checkCoreDeprecatedHooksInApiFiles` to `true` but I know Matt wants to do it in a separate minor release).
> 
> Commenting rather than explicitly approving because I don't know if we can already remove `checkCoreDeprecatedHooksInApiFiles` anywhere. For example from `bleedingEdge.neon`, in order to prevent people from getting deprecation notices for something they load from the package itself.
> 
> Not entirely sure what the difference in behaviour between `bleedingEdge.neon` and `extension.neon` is though (since the values also differ in those two files).

Good point about bleeding edge. Updated

src/Drupal/DrupalAutoloader.php
bbrala: I've updated bleedingedge so when people copy that they won't get a deprecation notice.

bbrala: Might need an isset or something when we remove the setting? Or perhaps even now.

mglaman: We don't need the `isset`, we can remove this when it is removed from `extension.neon` and the schema.

mglaman (approved):

PR #781, opened 2024-07-09
Add new class to exclusion [#780]

### LINKED ISSUE #780: InstallerExistingConfigTestBase is going to be deprecated
Issue: https://www.drupal.org/project/drupal/issues/3455820
Change Notice: https://www.drupal.org/node/3460001

Rule: https://github.com/mglaman/phpstan-drupal/blob/main/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php#L74

nlighteneddesign:
This should include the exclusion for the new class to match the now deprecated InstallerExistingConfigTestBase.

- fixes #780

### DIFF
--- a/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
+++ b/src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
@@ -71,6 +71,7 @@ final class BrowserTestBaseDefaultThemeRule implements Rule
 
         $excludedTestTypes = TypeCombinator::union(
             new ObjectType('Drupal\\FunctionalTests\\Update\\UpdatePathTestBase'),
+            new ObjectType('Drupal\\FunctionalTests\\Installer\\InstallerConfigDirectoryTestBase'),
             new ObjectType('Drupal\\FunctionalTests\\Installer\\InstallerExistingConfigTestBase')
         );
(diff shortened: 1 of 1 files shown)


src/Rules/Drupal/Tests/BrowserTestBaseDefaultThemeRule.php
```diff
@@ -71,6 +71,7 @@ public function processNode(Node $node, Scope $scope): array
 
         $excludedTestTypes = TypeCombinator::union(
             new ObjectType('Drupal\\FunctionalTests\\Update\\UpdatePathTestBase'),
+            new ObjectType('Drupal\\FunctionalTests\\Installer\\InstallerConfigDirectoryTestBase'),
             new ObjectType('Drupal\\FunctionalTests\\Installer\\InstallerExistingConfigTestBase')
         );
```
mglaman, line 76: We should only add it to the union if Drupal core is > 10.3.

nlighteneddesign: It's a brand new class, will it fail if it is in the union and it does not exist?

mglaman: It will not

Issue #364, opened 2022-03-12
datetime.module field type stubs


mglaman:
Part of https://github.com/mglaman/phpstan-drupal/issues/359. See [PHPStan documentation on Stub Files](https://phpstan.org/user-guide/stub-files).

Create stub files for the following field types in a `stubs/Drupal/datetime/Plugin/Field/FieldType` directory:

- `\Drupal\datetime\Plugin\Field\FieldType\DateTimeItem`

Add test cases to https://github.com/mglaman/phpstan-drupal/blob/main/tests/src/Type/data/field-types.php.

PR #785, opened 2024-07-11
Add stubs for datetime field type

ceesgeene:
Resolves https://github.com/mglaman/phpstan-drupal/issues/364

### DIFF
--- a/stubs/Drupal/datetime/Plugin/Field/FieldType/DateTimeItem.stub
+++ b/stubs/Drupal/datetime/Plugin/Field/FieldType/DateTimeItem.stub
@@ -1,0 +1,13 @@
+<?php
+
+namespace Drupal\datetime\Plugin\Field\FieldType;
+
+use Drupal\Core\Field\FieldItemBase;
+
+/**
+ * @property ?string $value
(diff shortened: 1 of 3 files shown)


stubs/Drupal/datetime/Plugin/Field/FieldType/DateTimeItem.stub
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Drupal\datetime\Plugin\Field\FieldType;
+
+use Drupal\Core\Field\FieldItemBase;
+
+/**
+ * @property ?string $value
+ * @property-read ?\Drupal\Core\Datetime\DrupalDateTime $date
```
mglaman, line 9: Technically it isn't read only. Core itself is doing 

```
$this->date = NULL;
```

Can we change this? Maybe when these phpdoc improvements go into core we can fix usages

ceesgeene: Sure thing!

mglaman (approved):
Thank you!!

Issue #369, opened 2022-03-12
datetime_range.module field type stubs


mglaman:
Part of https://github.com/mglaman/phpstan-drupal/issues/359

Create stub files for the following field types:

- `\Drupal\datetime_range\Plugin\Field\FieldType\DateRangeItem`

islam-m-ahmed:
i am intersted to do this work

PR #786, opened 2024-07-11
Daterange field type

ceesgeene:
https://github.com/mglaman/phpstan-drupal/issues/369

Note this MR is a follow up on https://github.com/mglaman/phpstan-drupal/pull/785

Also please carefully review the `@property-read null $date`. Not sure if this is the best solution. Maybe we should just remove the `extends DateTimeItem`.


stubs/Drupal/datetime_range/Plugin/Field/FieldType/DateRangeItem.stub
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Drupal\datetime_range\Plugin\Field\FieldType;
+
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItem;
+
+/**
+ * @property-read null $date
+ *   Even though this extends DateTimeItem, the date property does not actually
+ *   exist. It is not possible to undo a property definition.
```
mglaman, line 10: Oh... That is problematic.

mglaman (commented):
Need to see why the CI failed

PR #789, opened 2024-09-13
EntityReference formatter stubs

mstrelan:
Fixes the following issue (and other similar issues) when `checkFunctionArgumentTypes: true` is enabled

```
Parameter #1 $items of method Drupal\Core\Field\Plugin\Field\FieldFormatter\EntityReferenceFormatterBase::getEntitiesToView() expects Drupal\Core\Field\EntityReferenceFieldItemListInterface, Drupal\Core\Field\FieldItemListInterface given.
```


stubs/Drupal/Core/Field/Plugin/Field/FieldFormatter/EntityReferenceEntityFormatter.stub
mstrelan, line 1: It would be nice to add these just to `EntityReferenceFormatterBase` but that doesn't seem to work.

mstrelan:
Was this already fixed in https://github.com/mglaman/phpstan-drupal/pull/846?

mstrelan:
Closing as this is fixed. Confirmed by running phpstan against core with `checkFunctionArgumentTypes: true`. Previously a grep for "expects Drupal\Core\Field\EntityReferenceFieldItemListInterface" returned 13 errors, now returns 0.

PR #757, opened 2024-05-09
Adding the fix for the webflow 

DEVDESAGAR1:
Adding the fix for the webflow 

fixes #756

PR #790, opened 2024-09-24
Upgrade to webflo/drupal-finder 1.3

webflo:
Resolves https://github.com/mglaman/phpstan-drupal/pull/757

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -88,7 +88,7 @@ class DrupalAutoloader
         $drupalParams = $container->getParameter('drupal');
 
         // Trigger deprecation error if drupal_root is used.
-        if (isset($drupalParams['drupal_root'])) {
+        if (isset($drupalParams['drupal_root']) && $drupalParams['drupal_root'] === null) {
             trigger_error('The drupal_root parameter is deprecated. Remove it from your configuration. Drupal Root is discoverd automatically.', E_USER_DEPRECATED);
         }
 
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -91,7 +91,7 @@ class DrupalAutoloader
 
         $drupalRoot = $finder->getDrupalRoot();
         $drupalVendorRoot = $finder->getVendorDir();
-        if (! (bool) $drupalRoot || ! (bool) $drupalVendorRoot) {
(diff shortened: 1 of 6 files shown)


webflo:
@mglaman It should be possible to remove 'drupal_root' from the configuration completely.

mglaman:
We should be able to test dropping the `drupal_root` param with the package's own tests, as it relies on it

mglaman:
https://github.com/mglaman/phpstan-drupal/blob/main/tests%2Ffixtures%2Fconfig%2Fphpunit-drupal-phpstan.neon#L3

webflo:
@mglaman Should we decleare `drupal_root` as optional in the schema? It will break many configs without it.

mglaman:
Yeah let's define it as optional so nothing crashes

webflo:
@mglaman I could not figure out how to make it optional in the schema.

webflo:
I tried to use https://github.com/nette/schema/issues/27

webflo:
Lets set it do null initialy, I don't know how to use the deprecation method because it triggers a exception right away

mglaman:
This approach was used before https://github.com/mglaman/phpstan-drupal/blob/main/src%2FDrupal%2FDrupalAutoloader.php#L131-L134

I forgot we just added that. So manually triggering the error if the deprecated key exists

src/Drupal/DrupalAutoloader.php
```diff
@@ -84,19 +83,21 @@ class DrupalAutoloader
...
-         * @var array{drupal_root: string, bleedingEdge: array{checkDeprecatedHooksInApiFiles: bool, checkCoreDeprecatedHooksInApiFiles: bool, checkContribDeprecatedHooksInApiFiles: bool}} $drupalParams
+         * @var array{drupal_root: string|null, bleedingEdge: array{checkDeprecatedHooksInApiFiles: bool, checkCoreDeprecatedHooksInApiFiles: bool, checkContribDeprecatedHooksInApiFiles: bool}} $drupalParams
          */
         $drupalParams = $container->getParameter('drupal');
-        $drupalRoot = realpath($drupalParams['drupal_root']);
-        $finder = new DrupalFinder();
-        $finder->locateRoot($drupalRoot);
 
+        // Trigger deprecation error if drupal_root is used.
+        if (isset($drupalParams['drupal_root'])) {
```
mglaman, line 91: We should check if it's not `null`, since it's always set

mglaman:
Yay!

PR #803, opened 2024-09-27
Add rule identifiers

### LINKED ISSUE #772: Add error identifiers to phpstan-drupal rules.
# Feature request

_Apologies if this was already an opened issue. I tried searching first, but I'm not 100% confident I didn't miss something._

It appears that in PHPStan rules have now been given error identifiers in [v1.11](https://github.com/phpstan/phpstan/releases/tag/1.11.0). This provides users a way to ignore a specific rule if they want to based on the identifier.

The Drupal rules, however, do not output an identifier, and therefore cannot be ignored in the same manner as the parent project.

mglaman:
fixes #772


mglaman:
Fixed in https://github.com/mglaman/phpstan-drupal/pull/805

PR #805, opened 2024-11-11
Update phpstan/phpstan related dependencies to ^2.0"

Boegie:
Naive first stab

### DIFF
--- a/src/Rules/Classes/PluginManagerInspectionRule.php
+++ b/src/Rules/Classes/PluginManagerInspectionRule.php
@@ -123,7 +123,7 @@ class PluginManagerInspectionRule implements Rule
 
         if ($constructor->getDeclaringClass()->getName() !== $fqn) {
             $errors[] = RuleErrorBuilder::message(
-                $errors[] = sprintf('%s must override __construct if using YAML plugins.', $fqn)
+                sprintf('%s must override __construct if using YAML plugins.', $fqn)
             )
                 ->identifier('pluginManagerInspection.callAlterInfo')
                 ->build();
--- a/src/Rules/Classes/PluginManagerInspectionRule.php
+++ b/src/Rules/Classes/PluginManagerInspectionRule.php
@@ -8,11 +8,9 @@ use PhpParser\Node;
 use PhpParser\NodeFinder;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ReflectionProvider;
-use PHPStan\Rules\IdentifierRuleError;
 use PHPStan\Rules\Rule;
 use PHPStan\Rules\RuleErrorBuilder;
 use PHPStan\Type\ObjectType;
-use function sprintf;
 
 /**
  * @implements \PHPStan\Rules\Rule<\PhpParser\Node\Stmt\Class_>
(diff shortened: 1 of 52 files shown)


mglaman:
I'll probably want to open a 2.x branch as well. Thanks for trying at this.

Boegie:
No worries, I'll chip away at this for an hour 2 from now.

Boegie:
Ok, so this was more of a grind than actually Big Brain stuff.

I would recommend:

- Going over the identifiers that were (forcefully) added. There  are a lot of them and they might not be consistent.
- The changes in `.github/workflows/php.yml` can be reverted when drupal core includes `phpstan/phpstan:^2.0`
- Replacing all `@phpstan-ignore next-line` nuke-em-all lines with the more targeted `@phpstan-ignore [identifier]` in a separate MR. I found some "hidden sins" whilst creating this MR.

src/Rules/Classes/PluginManagerInspectionRule.php
```diff
@@ -78,9 +78,9 @@ public function processNode(Node $node, Scope $scope): array
             $errors[] = RuleErrorBuilder::message(
                 'Plugin managers should call alterInfo to allow plugin definitions to be altered.'
             )
-                ->identifier('plugin.manager.alterInfoMissing')
                 ->tip('For example, to invoke hook_mymodule_data_alter() call alterInfo with "mymodule_data".')
                 ->line($node->getStartLine())
+                ->identifier('pluginManagerInspection.alterInfoMissing')
                 ->build();
```
mglaman, line 84: I didn't review the commit log, but I'm guessing something about the format of the error identifier made PHPStan mad?

src/Rules/Drupal/GlobalDrupalDependencyInjectionRule.php
```diff
@@ -71,8 +72,9 @@ public function processNode(Node $node, Scope $scope): array
             return [];
         }
 
-        return [
-            '\Drupal calls should be avoided in classes, use dependency injection instead'
-        ];
+        return
+            [RuleErrorBuilder::message('\Drupal calls should be avoided in classes, use dependency injection instead')
+             ->identifier('globalDrupalDependencyInjectio.useDependencyInjection')
```
mglaman, line 77:
```suggestion
             ->identifier('globalDrupalDependencyInjection.useDependencyInjection')
```

mglaman (commented):
oh, did rule identifiers become required, essentially fixing https://github.com/mglaman/phpstan-drupal/pull/803 😱 I wish I could have landed that sooner to reduce this effort. I appreciate it. I literally fell asleep at my laptop 2 times while working on that.

mglaman:
> A new branch might be a nice clean cut-off point to drop D9 support?

We can say 2.x drops Drupal 9 support, as a follow up to his PR.

src/Rules/Classes/PluginManagerInspectionRule.php
Boegie: Nope, this is a rare occasion where there was actually an identifier and I completely overlooked it.

Not too sure what to do now...

Do we want double-full-stop identifiers? 

If we do I'll have to change the other identifiers in this class to the same prefix.

Help! :)

Boegie:
> I appreciate it. I literally fell asleep at my laptop 2 times while working on that.

I couldn't handle in batches longer than one hour, extremely grindy stuff

src/Rules/Classes/PluginManagerInspectionRule.php
mglaman: I'm fine taking your changes, I was just curious

mglaman:
Thanks for this!!

PR #807, opened 2024-11-15
Drop d9

### LINKED ISSUE #760: Stop testing PHP 7.4, 8.0 and Drupal 9
Managing this project would be much easier if PHP 7.4 and PHP 8.0 were dropped. This would also stop Drupal 9 testing; or a Drupal 9 job could be created with PHP 8.1.

webflo/drupal-finder was raised to PHP 8.1 in its 1.3 release. I'm afraid it'll get more complicated with D11, PHPUnit 10/11.

Proposal: 

- bump minimum PHP version to PHP 8.1
- drop testing D9
- drop lower versions of Drush, etc, which allowed D9 testing

this would trigger the 1.3.0 release

Boegie:
Closes #760


Boegie:
Do we wanna keep  `\mglaman\PHPStanDrupal\Rules\Deprecations\GetDeprecatedServiceRule` ? It's without tests now.

Boegie:
Also, does CircleCI work with anything else but D9? Looking at `.circleci/config.yml` I would say not?

If that's indeed the case, should we drop it entirely?

mglaman:
We can kill CircleCI, it's a left over

rules.neon
```diff
@@ -2,8 +2,6 @@ rules:
 	- mglaman\PHPStanDrupal\Rules\Drupal\Coder\DiscouragedFunctionsRule
 	- mglaman\PHPStanDrupal\Rules\Drupal\GlobalDrupalDependencyInjectionRule
 	- mglaman\PHPStanDrupal\Rules\Drupal\PluginManager\PluginManagerSetsCacheBackendRule
-	- mglaman\PHPStanDrupal\Rules\Deprecations\AccessDeprecatedConstant
```
mglaman, line 5: Do we need this rule in case future constants are deprecated? Or does PHP now support reading comments from constants... I think it does, so this rule isn't needed. Worth double checking.

mglaman (commented):
one thing I want to check

rules.neon
mglaman: Yeah it's been out for a while https://x.com/markusstaab/status/1765984217937154473

PR #817, opened 2025-01-15
Add support for shorthand service definition syntax

### LINKED ISSUE #816: Add support for shorthand service definition syntax
# Bug report

PHPStan with level 7 can't detect a combination of two services declared as FQNs, where one service references an interface that is used as an alias for another FQN service.

```yaml
services:
  _defaults:
    autowire: true

  Drupal\foo\Foo: {}
  Drupal\foo\FooInterface: '@Drupal\foo\Foo'
```

With a combination of `\Drupal\Core\DependencyInjection\ContainerInjectionInterface` it reports an error:

```php
  public function __construct(
    private FooInterface $foo,
  ) {}

  public static function create(ContainerInterface $container): self {
    return new self(
      $container->get(FooInterface::class),
    );
  }
```

```
Parameter #1 $foo of class <class> constructor expects Drupal\foo\FooInterface, object given.  
         🪪  argument.type
```

This happens, because in `\mglaman\PHPStanDrupal\Drupal\ServiceMap::setDrupalServices` there is no `$serviceDefinition` for such a service (its just an empty array). It goes to this section but fails because it's not a class but an interface:

https://github.com/mglaman/phpstan-drupal/blob/6fab2917c6ed26c1f515771c5ab2065448169016/src/Drupal/ServiceMap.php#L44-L46

It works with the `\Drupal\Core\Entity\EntityTypeManagerInterface`, because the alias `entity_type.manager` is a proper service definition with a `class` and `arguments`.

image

But fail with custom services using a shorthand definition, because there is no value for the class. It should be handled in the above code snippet, but only handles classes, and should also handle interfaces.

Niklan:
Closes #816

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -225,6 +225,16 @@ class DrupalAutoloader
                         }
                     });
                 }
+                // Handle shorthand syntax for service definition:
+                // @code
+                //   Drupal\foo\FooService: {}
+                //   Drupal\foo\BarService:
+                //     tags:
+                //       - { name: foo_bar }
+                // @endcode
+                if (!isset($serviceDefinition['class']) && class_exists($serviceId)) {
+                    $serviceDefinition['class'] = $serviceId;
+                }
                 // @todo sanitize "calls" and "configurator" and "factory"
                 /**
                 jsonapi.params.enhancer:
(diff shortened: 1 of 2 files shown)


mglaman (approved):

mglaman:
Thanks!

PR #823, opened 2025-01-24
Sync DrupalKernel synthetic services

### LINKED ISSUE #818: Handle DrupalKernel synthetic services
`DrupalKernel` registers 3 [synthetic services][synthetic-services] and instantly hardcodes aliases for two of them.

https://github.com/drupal/drupal/blob/c6bc6ebf27480b08b4c9ed5a7099ff5eef7e4348/core/lib/Drupal/Core/DrupalKernel.php#L1338-L1345

`mglaman\PHPStanDrupal\Drupal\DrupalAutoloader::register()` only provides information about services defined in the `*.services.yml` file, so those two services aren't detected.

This simple example triggers a problem:

```php
<?php

use Drupal\Core\DependencyInjection\ContainerInjectionInterface;
use Symfony\Component\DependencyInjection\ContainerInterface;

class Foo implements ContainerInjectionInterface {

  public function __construct(
    private ContainerInterface $container,
  ) {}

  public static function create(ContainerInterface $container): self {
    return new self(
      $container->get(ContainerInterface::class),
    );
  }

}
```

produces on PHPStan level 7:

```
  14     Parameter #1 $container of class Foo constructor expects Symfony\Component\DependencyInjection\ContainerInterface, object given.  
         🪪  argument.type 
```

If create is replaced by:

```
    return new self(
      $container,
    );
```

there is no error.

This is a kind of edge case scenario, but it raises a question: should this package handle it in some way? Considering that the services are not detected, this means that anything that is registered or changed by a service provider or compiler pass will not be detected. Perhaps it would be possible to improve things in general, rather than just this specific case?

[synthetic-services]: https://symfony.com/doc/current/service_container/synthetic_services.html

Niklan:
Closes #818 

This PR synchronizes current `DrupalKernel` synthetic service definitions with the `DrupalAutoloader`.

```php
    // Register synthetic services.
    $container->register('class_loader')->setSynthetic(TRUE);
    $container->register('kernel', 'Symfony\Component\HttpKernel\KernelInterface')->setSynthetic(TRUE);
    $container->register('service_container', 'Symfony\Component\DependencyInjection\ContainerInterface')->setSynthetic(TRUE);

    // Register aliases of synthetic services for autowiring.
    $container->setAlias(DrupalKernelInterface::class, 'kernel');
    $container->setAlias(ContainerInterface::class, 'service_container');
```

https://github.com/drupal/drupal/blob/c6bc6ebf27480b08b4c9ed5a7099ff5eef7e4348/core/lib/Drupal/Core/DrupalKernel.php#L1338-L1345


src/Drupal/DrupalAutoloader.php
```diff
@@ -108,7 +109,10 @@ public function register(Container $container): void
         // Attach synthetic services
         // @see \Drupal\Core\DrupalKernel::attachSynthetic
         $this->serviceMap['kernel'] = ['class' => DrupalKernelInterface::class];
+        $this->serviceMap[DrupalKernelInterface::class] = ['alias' => 'kernel'];
         $this->serviceMap['class_loader'] = ['class' => ClassLoader::class];
+        $this->serviceMap['service_container'] = ['class' => \Drupal\Component\DependencyInjection\Container::class];
```
Niklan, line 114: I'm not sure if this should be aliased to `DrupalContainer`?

Niklan: I guess the linter tells itself. :smile:

mglaman (approved):

PR #826, opened 2025-02-16
Add type specifier for Drupal\Component\Assertion\Inspector

Niklan:
This Pull Request adds a type specifier for `Drupal\Component\Assertion\Inspector`.

## Problem/motivation

Drupal provides the `Drupal\Component\Assertion\Inspector` class, which makes it easier and more readable to assert in some cases, for example:

```php
\assert(\is_string($form_state->getValue(['foo'])));
\assert(\is_string($form_state->getValue(['baz'])));
\assert(\is_string($form_state->getValue(['baz'])));
```

can be improved using this class:

```php
\assert(Inspector::assertAllStrings([
  $form_state->getValue(['foo']),
  $form_state->getValue(['bar']),
  $form_state->getValue(['baz']),
]));
```

Reading such code is much easier visually.

This PR clarifies how PHPStan should infer types following an assert, avoiding the use of “mixed” in the code.

P.S. This is my first extension for PHPStan, so it might be wrong in some ways.


Niklan:
The last CI failure seems like a false positive to me.

PR #846, opened 2025-03-29
Make FormatterInterface stub generic

Niklan:
From #822: This PR changes FormatterStub to be a proper generic instead of narrowing types for others.

**Conflict with #844.**


Niklan:
@ceesgeene, if possible, please take a look and give feedback on the generic proposal.

Niklan:
The only problem I faced is that I can't define these generic subtypes using a docblock above the class.

```php
/**
 * @implements FormatterInterface<CommentFieldItemList>
 */
```

This leads to an error: `Class {class} has @implements tag, but does not implement any interface`, because it actually extends `FormatterBase`, which implements the `FormatterInterface`. This appears to make it impossible to declare generics `TFieldItemList` for both methods at the same time:

```
Method {formatter}::viewElements() has parameter $items with generic interface Drupal\Core\Field\FieldItemListInterface but does not specify its types: T
```

This forces us to add types directly at the method level:

```php
/**
 * @param CommentFieldItemList $items
 */
public function viewElements(FieldItemListInterface $items, $langcode): array {
```

Then it works. But in that case, it's unclear what the generic is actually doing. You can simply remove it and get the same result.

So it appears that we also should provide stub for `FormatterBase`:

```php
<?php

namespace Drupal\Core\Field;

use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
use Drupal\Core\Field\FormatterInterface;
use Drupal\Core\Field\PluginSettingsBase;

/**
 * @template TFieldItemList of \Drupal\Core\Field\FieldItemListInterface
 * @implements FormatterInterface<TFieldItemList>
 */
abstract class FormatterBase extends PluginSettingsBase implements FormatterInterface, ContainerFactoryPluginInterface {

}
```

In that case it is possible to tell PHPStan what types we are using:

```php
/**
 * @extends FormatterBase<CommentFieldItemList>`
 */
final class FooFormatter extends FormatterBase {
```

Niklan:
@mglaman how to handle these test failures? It means we also should add stubs for `ContainerFactoryPluginInterface` and `PluginSettingsBase` or add them into baseline?

mglaman:
@Niklan it means we need the missing stubs added. Making a stub requires providing stubs for any extends or implementation, even if they're empty. See https://github.com/mglaman/phpstan-drupal/blob/main/stubs/Drupal/Component/Plugin/PluginInspectionInterface.stub as an example

Niklan:
@mglaman this one and #847 are ready for review. I think there is nothing I can do about test failures in HEAD baseline check action.

stubs/Drupal/Core/Field/FormatterInterface.stub
```diff
@@ -2,14 +2,25 @@
...
 interface FormatterInterface {
 
   /**
-   * @param \Drupal\Core\Field\FieldItemListInterface<\Drupal\Core\Field\FieldItemInterface> $items
+   * @param TFieldItemList $items
    * @param string $langcode
    *
    * @return array<int, array<int|string, mixed>>
    */
   public function viewElements(FieldItemListInterface $items, $langcode): array;
```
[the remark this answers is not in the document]
mglaman: `MissingMethodParameterTypehintRule` returns the error, we could make a test which calls it to see if this fixes. But there _should_ be some kind of type test we can write instead.

mglaman:
I pushed [`e158c6d` (#846)](https://github.com/mglaman/phpstan-drupal/pull/846/commits/e158c6d83a4051f1eba15184db9fdbdb001af3c4) as I think it's the correct way to pass the template down to the field item list and provide the generic for formatters

mglaman:
Ah I overlooked this example:

```
/**
 * @extends FormatterBase<FieldItemList<StringItem>>
 */
final class FooFormatter extends FormatterBase {
```

So maybe it was the same all along and it comes down to style of writing generics?

stubs/Drupal/Core/Field/FormatterInterface.stub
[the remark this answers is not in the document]
Niklan: I don't think the template name makes any difference; it's just a preference or an alias, as I understood it. It won't pass it to other inherited generics automatically anyway.

The reason I gave it such an explicit alias is that `FormatterInterface` works not only with `FieldItemListInterface`. There is a chance that in the future `FieldDefinitionInterface` or `FormStateInterface` (form object can be specified through it) might also require a generic template. In this case, `TFieldItemList` will be self-documenting, while `T` might cause some confusion. 

I don't mind changing it to `T`, as PHPStan allows changing that name freely; the only thing we should care about for BC is the order of templates. So when we need a second one, we can make the name clearer.

I fully agree about tests; I will try to create one.

stubs/Drupal/Core/Field/FormatterInterface.stub
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @template T of FieldItemInterface
+ */
+interface FormatterInterface extends PluginSettingsInterface {
+
+  /**
+   * @param array<\Drupal\Core\Field\FieldItemListInterface<T>> $entities_items
+   */
+  public function prepareView(array $entities_items): void;
```
Niklan, line 15: @mglaman I've pushed an initial test for generic. I think we should also cover `FormatterBase` in that case, but it's later. It will fail and highlight the actual problem.

With this type of generic, **you can't specify the list items' type**, which should be possible (see `BazFormatter` from test).

For example, my formatter is working with `FileItem` field items, which is an entity reference field item. It even has its own list `FileFieldItemList`, so I must explicitly set it:

```php
/**
 * @implements FormatterInterface<FileFieldItemList<FileItem>>
 */
```

This is very important because `FileFieldItemList` can contain additional methods, overrides, and other stuff that can affect the actual data. One of such methods is `::referencedEntities()` inherited from `EntityReferenceFieldItemList`. With the current generic, PHPStan wouldn't be able to understand and "see" this method because it always thinks it's `FieldItemListInterface<FileItem>`, which is technically true but is a very wide type that loses a lot of additional information.

I think my initial generic was correct:

```php
/**
 * @template T of \Drupal\Core\Field\FieldItemListInterface
 */
interface FormatterInterface {
```

```php
  /**
   * @param T $items
   */
  public function view(FieldItemListInterface $items, $langcode = NULL);
```

Take a closer look at it. It forces you to specify the **type of item list** class because this is what the formatter parameters are expecting (`FieldItemListInterface $items`). The item list is already a generic, so you can specify their items as well: `FormatterInterface<FieldItemList<StringItem>>`. This way, PHPStan will understand that `$items` is `FieldItemList` and contains items of type `StringItem`. So any code from `FieldItemList` will be visible to PHPStan.

This can be overwhelming to specify two types at the same time, but this is how it should work. With the file field example, it will be `FormatterInterface<FileFieldItemList<FileItem>>`, but since `FileFieldItemList` is kind of a specific solution, the class itself should have the item specification:

```php
/**
 * @extends FieldItemList<FileItem>
 */
class FileFieldItemList ...
```

In that case, if this is defined at the `FileFieldItemList` level, the formatter can just provide `FormatterInterface<FileFieldItemList>`, which is equivalent to `FormatterInterface<FileFieldItemList<FileItem>>`, but this time, information about `FileItem` is obtained from the list itself.

---

Maybe I'm wrong, but I think this is the proper solution because now you can only define the item type but not the item list, which the formatter is actually expecting. To put it simply: the formatter is working with a field item list contract, not directly with the field item. It is the item list that dictates what kind of items it contains (`FileFieldItemList` → `FileItem`), not the formatter.

Niklan: Oh, I didn't saw you provided tests as well >< Anyway, lets discuss it. We can remove mine.

ceesgeene:
Note that it is possible to provide template defaults since https://github.com/phpstan/phpstan-src/pull/3457

The FormatterInterface could have two templates l, one for the field item and one for the list, were the one for the list is optional and defaults to FieldItemListInterface

```
/**
 * @template TFieldItem
 * @template TFieldItemList = FieldItemListInterface 
 */
interface FormatterInterface
```

Formatters then could only provide the class for the TFieldItem template if they don't need methods of a more specific field item list interface.

Not sure if I would use this feature though. E.g. I don't know how well this feature is supported by IDEs.

stubs/Drupal/Core/Field/FormatterInterface.stub
mglaman: Thanks for the write-up, going to make sure I set aside time to read again. But I do agree - I think your approach is correct, and what I proposed is not.

mglaman:
I'm going to revert the commits I made and merge the original

Niklan:
@mglaman I think your approach to testing generic is much better than mine. Maybe we should try to improve the test to cover more cases, and revert the generic definition back to its initial state. Since the problem version of the generic has been reverted and released, we can do this properly without rushing anything.

Niklan:
@mglaman I've removed my tests, reverted the generic definition to its initial state (but with the changes you've made to `::prepareView()`). I'll try to improve the tests.

Btw, how should the generic template be called? I've reverted it to `TFieldItemList`. As mentioned already, it doesn't matter; we can freely change its name. The only thing we should care about is the order of template notations when multiple data become generics. So, I can make it `T` or keep it as `TFieldItemList`.

The only downside to changing the name of the generic is that if someone adds it into the baseline or `ignoreErrors` using regular expressions, it would fail then.

Niklan:
@mglaman also, I have a question regarding the `formatter-interface.php`. Any reason why you've added methods required by `FormatterInterface`?

For example, this one:

```php
/**
 * @implements \Drupal\Core\Field\FormatterInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem>
 */
class StringFormatter implements FormatterInterface {

    public function settingsForm(array $form, FormStateInterface $form_state) {
        return [];
    }

    public function settingsSummary() {
        return [];
    }

    public function prepareView(array $entities_items) {
        $first_item = $entities_items[0]->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
    }

    public function view(FieldItemListInterface $items, $langcode = NULL) {
        assertType('Drupal\Core\Field\FieldItemListInterface<Drupal\Core\Field\Plugin\Field\FieldType\StringItem>', $items);
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }

    public function viewElements(FieldItemListInterface $items, $langcode) {
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }

    public static function isApplicable(FieldDefinitionInterface $field_definition) {
        return true;
    }

    public function getSettings() {
        return [];
    }

    public function getSetting($key) {
        return null;
    }

    public function setSetting($key, $value) {
        return $this;
    }

    public function setSettings(array $settings) {
        return $this;
    }

    public function getDefaultSettings() {
        return [];
    }

    public function getPluginId() {
        return 'string_formatter';
    }

    public function getPluginDefinition() {
        return [];
    }

    public function getThirdPartySettings($provider = NULL) {
        return [];
    }

    public function getThirdPartySetting($provider, $key, $default = NULL) {
        return $default;
    }

    public function setThirdPartySetting($provider, $key, $value) {
        return $this;
    }

    public function unsetThirdPartySetting($provider, $key) {
        return $this;
    }

    public function getThirdPartyProviders() {
        return [];
    }
}
```

could be simplified to

```php
/**
 * @implements \Drupal\Core\Field\FormatterInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem>
 */
class StringFormatter implements FormatterInterface {

    public function prepareView(array $entities_items) {
        $first_item = $entities_items[0]->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
    }

    public function view(FieldItemListInterface $items, $langcode = NULL) {
        assertType('Drupal\Core\Field\FieldItemListInterface<Drupal\Core\Field\Plugin\Field\FieldType\StringItem>', $items);
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }

    public function viewElements(FieldItemListInterface $items, $langcode) {
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }
    
}
```

Yes, it's wrong from the PHP interpreter (and LSP) point of view because the class does not implement required methods. But it's not real PHP code; PHPStan doesn't care about it since all `**/data/*` files are explicitly excluded. Tests work properly that way because PHPStan is a static analyzer; it doesn't execute that file, so its syntax validity is not important.

But doing it without "empty" methods just to please PHP LSP makes tests much smaller and clearer about what is actually tested. These empty and useless methods add too much noise and distract from what is actually tested.

If I remove all such methods, the tests become so much smaller, like literally 238 lines of code to 111, basically twice as small, doing exactly the same thing.

```php
<?php

namespace DrupalFormatterInterface;

use Drupal\Core\Field\FieldItemListInterface;
use Drupal\Core\Field\FormatterBase;
use Drupal\Core\Field\FormatterInterface;
use function PHPStan\Testing\assertType;

/**
 * @implements \Drupal\Core\Field\FormatterInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem>
 */
class StringFormatter implements FormatterInterface {

    public function prepareView(array $entities_items) {
        $first_item = $entities_items[0]->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
    }

    public function view(FieldItemListInterface $items, $langcode = NULL) {
        assertType('Drupal\Core\Field\FieldItemListInterface<Drupal\Core\Field\Plugin\Field\FieldType\StringItem>', $items);
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }

    public function viewElements(FieldItemListInterface $items, $langcode) {
        $first_item = $items->first();
        // Assert that the first item in the list is a StringItem
        assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $first_item);
        return [];
    }

}

/**
 * @template T of \Drupal\Core\Field\FieldItemInterface
 * @extends \Drupal\Core\Field\FormatterBase<T>
 */
class GenericExtendsFormatterBase extends FormatterBase {
    /**
     * @param \Drupal\Core\Field\FieldItemListInterface<T> $items
     */
    public function viewElements(FieldItemListInterface $items, $langcode) {
        $first = $items->first();
        assertType('T of Drupal\Core\Field\FieldItemInterface (class DrupalFormatterInterface\GenericExtendsFormatterBase, argument)|null', $first);
        return [];
    }
}

/**
 * @template T of \Drupal\Core\Field\FieldItemInterface
 */
class GenericFormatter implements FormatterInterface {

    /**
     * @param \Drupal\Core\Field\FieldItemListInterface<T> $items
     */
    public function view(FieldItemListInterface $items, $langcode = NULL) {
        $first = $items->first();
        assertType('T of Drupal\Core\Field\FieldItemInterface (class DrupalFormatterInterface\GenericFormatter, argument)|null', $first);
        return [];
    }

    /**
     * @param \Drupal\Core\Field\FieldItemListInterface<T> $items
     */
    public function viewElements(FieldItemListInterface $items, $langcode) {
        $first = $items->first();
        assertType('T of Drupal\Core\Field\FieldItemInterface (class DrupalFormatterInterface\GenericFormatter, argument)|null', $first);
        return [];
    }
}

/**
 * Test usage of the formatter through a function.
 *
 * @param \Drupal\Core\Field\FormatterInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem> $formatter
 * @param \Drupal\Core\Field\FieldItemListInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem> $items
 */
function testFormatterWithStringItems(FormatterInterface $formatter, FieldItemListInterface $items): void {
    $elements = $formatter->viewElements($items, 'en');
    assertType('array<int, array<int|string, mixed>>', $elements);

    // Test with implementation
    $stringFormatter = new StringFormatter();
    $elements = $stringFormatter->viewElements($items, 'en');
    assertType('array<int, array<int|string, mixed>>', $elements);

    $item = $items->first();
    // The generic type from FormatterInterface should flow to $item
    assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $item);
}

/**
 * Test that FormatterBase correctly implements FormatterInterface with generics.
 *
 * @param \Drupal\Core\Field\FormatterBase<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem> $formatter
 * @param \Drupal\Core\Field\FieldItemListInterface<\Drupal\Core\Field\Plugin\Field\FieldType\StringItem> $items
 */
function testFormatterBaseWithStringItems(FormatterBase $formatter, FieldItemListInterface $items): void {
    $output = $formatter->view($items, 'en');
    assertType('array<int|string, mixed>', $output);

    $item = $items->first();
    // The generic type from FormatterBase should flow to $item
    assertType('Drupal\Core\Field\Plugin\Field\FieldType\StringItem|null', $item);
}
```

I want your input on this. I won't delete anything for now. But what do you think about it?

mglaman:
Yeah it can be trimmed up. That was generated by Copilot for the data source and I didn't clean it up. Feel free to undo and change anything I added.

I've got the same error you reported on my blog code base and will be testing your fix that way

Niklan:
@mglaman I've updated the tests, and the last two of them are failing: `ExtendedDeepFormatter` and `ExtendsBooleanItemFormatter`.

They are failing very weirdly, for example, this one:

```php
/**
 * @extends FormatterBase<FakeBooleanFieldItemList>
 */
class ExtendsBooleanItemFormatter extends FormatterBase {

    public function prepareView(array $entities_items): void {
        assertType('array<DrupalFormatterInterface\FakeBooleanFieldItemList>', $entities_items);
```

Fails with an error:

```
--- Expected
+++ Actual
@@ @@
-'array<DrupalFormatterInterface\FakeBooleanFieldItemList>'
+'array<Drupal\Core\Field\TFieldItemList>'
```

For some reason, it resolves `TFieldItemList` as a some kind of class/interface. Even if I replace it with `T`, it reports `array<Drupal\Core\Field\T>` (FQN doesn't helps here as well).

Another weird part is that it doesn't report problems in `::viewElements()` for both classes, which is odd, because they have the same generic interface and PHPDoc, exactly the same asserts, but they are not failing, while `::view()` and `::prepareView()` are failing.

Maybe we faced this bug: https://github.com/phpstan/phpstan/issues/9708, but I'm unsure, because it should report `::viewElements()` as well then.

P.S. I have a feeling I'm doing tests wrong for that case. PHPStan itself does it completely different: https://github.com/phpstan/phpstan-src/tree/2.1.x/tests/PHPStan/Generics/data. I'm also thinking we should have such tests under the `Generics` namespace to clearly separate them from `Type`.

mglaman:
I'm all for moving to a Generics namespace for the data to keep it isolated

Niklan:
@mglaman will move them next time. :+1: 

I found a pattern: when extending `FormatterBase`, the `::prepareView()` and `::view()` methods actually have overrides in the Drupal implementation, but `::viewElements()` does not. Maybe it leads to conflicts in some way, or it is a bug in PHPStan.

Making a stub that looks like this helps:

```php
<?php

namespace Drupal\Core\Field;

use Drupal\Core\Plugin\ContainerFactoryPluginInterface;

/**
 * @template TFieldItemList of Drupal\Core\Field\FieldItemListInterface
 * @implements FormatterInterface<TFieldItemList>
 */
abstract class FormatterBase extends PluginSettingsBase implements FormatterInterface, ContainerFactoryPluginInterface {

  /**
   * @param TFieldItemList $items
   */
  public function view(FieldItemListInterface $items, $langcode = NULL) {}

  /**
   * @param array<TFieldItemList> $items
   */
  public function prepareView(array $entities_items) {}

}
```

But I don't like it. It basically duplicates information from the `FormatterInterface` partially, which looks weird.

As I understand it: the stub doesn't define these methods in any way → the original class has implementations of them, which takes priority over interface → PHPStan can't figure out the generic types in the PHPDoc for these methods.

Original overrides are simple:

```php
  /**
   * {@inheritdoc}
   */
  public function prepareView(array $entities_items) {}
  
  /**
   * {@inheritdoc}
   */
  public function prepareView(array $entities_items) {}
```

Assumption: this `{@inheritdoc}` overrides interface `@param array<TFieldItemList> $entities_items` which leads to a problem. Looks like PHPStan doesn't check interface PHPDoc in that case, at least for generics.

mglaman:
@Niklan I think we're just going to have to be extra verbose in phpstan-drupal with our stubs until we can fix core, unfortunately. So basically we're going to be over-stubbing to avoid any confusion in PHPStan, correct?

Niklan:
@mglaman sounds correct, simply because with overrides Drupal technically can "break" generics by forcing/reset the other types, but it should be detected in advance, so it's not a major problem. But yes, looks like we're going to stub much more than needed, because for now, it looks like the order how PHPStan is resolving generic is: stub → original → parent (inherited, class/interface) stubs → originals → repeat. If some data is missing in one, it looks it in the other. `FormatterBase` stub does not have overrides (for now) `::view()` and `::prepareView()`, so it gets the data from original class, which has overrides with `{@inheritdocs}`, and I assume it doesn't check the interface because of it, like it treats "PHPDoc as certain". But this PHPDoc doesn't define parameters, so it is kind of resets them to typehints of the method (in that case test errors makes sense and they are correct). But on the other hand, `::viewElements()` is not reporting any errors, because it is not overridden in original `FormatterBase`.

It all sounds logical, but the problem is that I can't reproduce it with PHPStan playground: https://phpstan.org/r/03d78262-0b50-469a-99dc-0ef4e1ee37f5. `BazFormatter::viewFoo()` should actually fail here if my assumption is correct, but it doesn't.

On the other hand, adding methods to the `FormatterBase` stub in our case solves the problem, so we can move forward. I still wish to understand why we have such a problem and Playground doesn't. The only difference is that there are no stubs in the playground that could cause issues. I'll try to dig more into that, maybe I'll be able to find out why it's failing. I prefer not to stub methods which should be properly inherited from other stubs/original classes/interfaces and doesn't require overrides for PHPStan.

Niklan:
This one looks like a bug:

https://github.com/mglaman/phpstan-drupal/blob/be02a3c69e00c221abf787d95bbaac40240247a9/stubs/Drupal/Core/Field/FieldItemList.stub#L7-L16

Shouldn't it be:

```php
  /**
   * @return T
   */
```

tests/src/Generics/data/formatter-interface.php
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * @implements FormatterInterface<BooleanFieldItemList>
+ */
+class BooleanItemFormatter implements FormatterInterface {
+
+    public function prepareView(array $entities_items): void {
+        assertType('array<DrupalFormatterInterfaceGeneric\BooleanFieldItemList>', $entities_items);
+        $items = $entities_items[0];
+        assertType('DrupalFormatterInterfaceGeneric\BooleanFieldItemList', $items);
+        assertType('Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem|null', $items->first());
```
Niklan, line 193: I need a fresh look on this. I don't understand why:

```
assertType('Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem|null', $items->first());
```

fails with an error:

```
1) mglaman\PHPStanDrupal\Tests\Generics\FormatterInterfaceGenericTest::testFileAsserts with data set "/var/www/html/tests/src/Generics/data/formatter-interface.php:193" ('type', '/var/www/html/tests/src/Gener...ce.php', 'Drupal\Core\Field\Plugin\Fiel...m|null', '*ERROR*', 193)
Expected type Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem|null, got type *ERROR* in /var/www/html/tests/src/Generics/data/formatter-interface.php on line 193.
--- Expected
+++ Actual
@@ @@
-'Drupal\Core\Field\Plugin\Field\FieldType\BooleanItem|null'
+'*ERROR*'
```

It clearly properly resolves the `$items` type, but why does it fail to resolve the item type?

mglaman: I did a PhpStorm look up for new instance creation of `ErrorType` and saw:

```
    $typeMap = \PHPStan\Type\Generic\TemplateTypeMap::createEmpty();
        foreach ($this->getTypes() as $i => $type) {
            $other = $otherTypes[$i] ?? new ErrorType();
```

So the `$otherTypes` does not have that type for some reason

Niklan: For reference, it appears to be working in PHPStan properly: https://phpstan.org/r/7dc786ec-c90d-46e9-b8f7-9d6d1172a951 so its should be doable

Niklan: Another reference provided by @mglaman in Slack: https://phpstan.org/r/89c23c16-7467-4411-8f81-489809e5ee52 which is also works as expected

Niklan: For those who found it and faced the same symptoms: the fix was in properly adding the directory with test cases (data) into composer.json → autoload-dev → classmap.

mglaman:
I'm clicking "update branch" so we can get a granular check on the 11.x-dev fail – to see if it's 11.0, 11.1, 11.2.x-dev or stuff for 12.x

mglaman:
💀 so it was something in 11.x

stubs/Drupal/Core/Field/FieldItemList.stub
```diff
@@ -5,16 +5,15 @@ namespace Drupal\Core\Field;
...
  */
 class FieldItemList extends ItemList implements FieldItemListInterface {
 
   /**
-   * @return \Drupal\Core\Field\FieldItemInterface
+   * @return T
    */
-  protected function createItem(int $offset = 0, ?mixed $value = NULL): \Drupal\Core\Field\FieldItemInterface {
-  }
+  protected function createItem(int $offset = 0, mixed $value = NULL): FieldItemInterface {}
```
mglaman, line 17: Technically it needs to stay `?mixed`? or does `mixed` imply null?

stubs/Drupal/Core/TypedData/Plugin/DataType/ItemList.stub
```diff
@@ -13,4 +13,10 @@ use Drupal\Core\TypedData\TypedDataInterface;
  * @implements ListInterface<T>
  */
 class ItemList extends TypedData implements \IteratorAggregate, ListInterface {
+
+  /**
+   * @return \ArrayIterator<int, T>
+   */
+  public function getIterator(): \ArrayIterator {}
```
mglaman, line 20: I wonder if we should split this out into its own PR, just show it is more apparent. It was crucial to fixing this PR and removes ignores errors in Drupal core's baseline. 

"Stub ItemList::getIterator return type" and then the PR can contain your findings

mglaman (commented):
One thought, sorry. I just think the "small" change in ItemList stub is worth it's own PR release note line

stubs/Drupal/Core/Field/FieldItemList.stub
Niklan: `?mixed` is wrong from the syntax point of view, as it implies `NULL`. I thought it could lead to some issues because this could lead to a wrong AST. But it clearly wasn't an issue, so I can revert it back, no problem. However, it would be better to consider removing the nullable type.

https://www.php.net/manual/en/language.types.mixed.php: 

> The [mixed](https://www.php.net/manual/en/language.types.mixed.php) type accepts every value. It is equivalent to the [union type](https://www.php.net/manual/en/language.types.type-system.php#language.types.type-system.composite.union) `object|resource|array|string|float|int|bool|null`. Available as of PHP 8.0.0.

stubs/Drupal/Core/TypedData/Plugin/DataType/ItemList.stub
Niklan: No problem, I can prepare PRs with this change. Should this one: https://github.com/mglaman/phpstan-drupal/pull/846#issuecomment-2790236394 be part of new PR? Because they are kind of connected. Or should I keep it with this PR?

Niklan:
Converting both PRs to draft to work on: https://github.com/mglaman/phpstan-drupal/pull/846#issuecomment-2787188862

PR #853, opened 2025-04-09
Fixes for InspectorTypeExtension

### LINKED ISSUE #852: Inspector type specifier breaks subsequent calls after Inspector
# Bug report

See https://www.drupal.org/project/project_browser/issues/3518231

```php
   * @param array $images
   *   Images of the project. Each item needs to be an array with two elements:
   *   `file`, which is a \Drupal\Core\Url object pointing to the image, and
   *   `alt`, which is the alt text.
```

```php
    assert(
      Inspector::assertAllArrays($images) &&
      Inspector::assertAll(fn (array $i): bool => $i['file'] instanceof Url, $images) &&
      Inspector::assertAllHaveKey($images, 'alt')
    ) or throw new \InvalidArgumentException('The project images must be arrays with `file` and `alt` elements.');
```

https://git.drupalcode.org/project/project_browser/-/blob/2.0.x/src/ProjectBrowser/Project.php?ref_type=heads#L63

mglaman:
fixes #852

### DIFF
--- a/src/Type/InspectorTypeExtension.php
+++ b/src/Type/InspectorTypeExtension.php
@@ -109,7 +109,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
 
         return $this->typeSpecifier->create(
             $node->getArgs()[1]->value,
-            new IterableType(new MixedType(true), new MixedType(true)),
+            new IterableType(new MixedType(), new MixedType()),
             TypeSpecifierContext::createTruthy(),
             $scope,
         );
@@ -122,7 +122,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
     {
         return $this->typeSpecifier->create(
             $node->getArgs()[0]->value,
-            new IterableType(new MixedType(true), new StringType()),
+            new IterableType(new MixedType(), new StringType()),
             TypeSpecifierContext::createTruthy(),
             $scope,
         );
@@ -135,7 +135,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
     {
         // Drupal considers string as part of "stringable" as well.
         $stringable = TypeCombinator::union(new ObjectType(Stringable::class), new StringType());
-        $newType = new IterableType(new MixedType(true), $stringable);
+        $newType = new IterableType(new MixedType(), $stringable);
 
         return $this->typeSpecifier->create(
             $node->getArgs()[0]->value,
@@ -150,8 +150,8 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
      */
     private function specifyAssertAllArrays(MethodReflection $staticMethodReflection, StaticCall $node, Scope $scope, TypeSpecifierContext $context): SpecifiedTypes
     {
-        $arrayType = new ArrayType(new MixedType(true), new MixedType(true));
-        $newType = new IterableType(new MixedType(true), $arrayType);
+        $arrayType = new ArrayType(new MixedType(), new MixedType());
+        $newType = new IterableType(new MixedType(), $arrayType);
 
         return $this->typeSpecifier->create(
             $node->getArgs()[0]->value,
@@ -170,7 +170,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
             // In Drupal, 'strict arrays' are defined as arrays whose indexes
             // consist of integers that are equal to or greater than 0.
             IntegerRangeType::createAllGreaterThanOrEqualTo(0),
-            new MixedType(true),
+            new MixedType(),
         );
 
         return $this->typeSpecifier->create(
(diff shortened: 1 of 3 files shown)


src/Type/InspectorTypeExtension.php
```diff
@@ -104,13 +103,13 @@ private function specifyAssertAll(MethodReflection $staticMethodReflection, Stat
...
 
-        if (!$callableInfo instanceof ClosureType) {
+        if (!$callableInfo->isCallable()->yes()) {
             return new SpecifiedTypes();
         }
 
         return $this->typeSpecifier->create(
             $node->getArgs()[1]->value,
-            new IterableType(new MixedType(true), $callableInfo->getReturnType()),
+            new IterableType(new MixedType(true), new MixedType()),
```
mglaman, line 112: Should this be:

```suggestion
            new IterableType(new MixedType(true), new MixedType(true)),
```

src/Type/InspectorTypeExtension.php
```diff
@@ -234,9 +233,9 @@ private function specifyAssertAllHaveKey(MethodReflection $staticMethodReflectio
             $keyTypes[] = new HasOffsetType(new ConstantStringType($key));
         }
 
-        $newArrayType = new ArrayType(
+        $newArrayType = new IterableType(
             new MixedType(true),
-            new ArrayType(TypeCombinator::intersect(new MixedType(), ...$keyTypes), new MixedType(true)),
+            new ArrayType(TypeCombinator::union(new MixedType(), ...$keyTypes), new MixedType(true)),
```
mglaman, line 238: Intersect gives us `hasOffset` but it's lost in `union`... maybe because we keep `MixedType` or `MixedType` is not `true`? If we have `HasOffsetType` of string... can we even have `MixedType` anymore?

So 

```
$keyTypes[] = new IntegerType();
$keyTypes[] = new StringType();

new ArrayType(TypeCombinator:: intersect(...$keyTypes), new MixedType(true)),
```

tests/src/Type/data/bug-852.php
```diff
@@ -0,0 +1,23 @@
...
+
+    assert(Inspector::assertAllHaveKey($baz, 'alt'));
+    assertType('array<mixed, array<mixed, mixed>>', $baz);
+}
+
+function bar(array $zed): void {
+    assert(Inspector::assertAll(fn (string $value) => $value === 'foo', $zed));
+    assertType('array<mixed, string>', $zed);
+
+}
```
mglaman, line 23: I'd like this to prove `hasOffset`, but we could merge and make it a follow up.

src/Type/InspectorTypeExtension.php
mglaman: Removing `MixedType` still causes `'*NEVER*`

src/Type/InspectorTypeExtension.php
Niklan:
```suggestion
            new IterableType(new MixedType(), new MixedType()),
```

I think it should be this. I'm not sure exactly, because this was my first (and only for now) extension, but it looks like the only difference between explicit Mixed and implicit is the use of that value "as boolean" (`\PHPStan\Rules\BooleansInConditions\BooleanRuleHelper::passesAsBoolean`).

If I understand it correctly, it reads as: if the mixed type is not marked as explicit, then it can be used as a boolean indicator (at users risk).

```php
$mixed_var = something_that_returns_mixed();

// This is allowed if NOT explicit.
// But if it IS explicit, it is forbidden to use as boolean?
if ($mixed_var) {}
```

So, the use of explicit mixed values in conditions should lead to an error:

```
Only booleans are allowed in an if condition, %s given.
```

I've did a very dirty check `src/test.php` → `phpstan analyze ./src/test.php` (temporary added `extension.neon` into `includes`):

```php
<?php

function some_array(): array {return [];}

$var = some_array();
\Drupal\Component\Assertion\Inspector::assertAll(fn (): bool => TRUE, $var);
foreach ($var as $key => $value) {
    if ($key) {}
    if ($value) {}
}
```

and got this results:


| TKey                  | TValue                | Analyze result                                                                                                                                                                                     |
| --------------------- | --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `new MixedType()`     | `new MixedType()`     | `Only booleans are allowed in an if condition, (int\|string) given.` reported for `if ($key) {}`                                                                                                   |
| `new MixedType(true)` | `new MixedType()`     | `Only booleans are allowed in an if condition, (int\|string) given.` reported for `if ($key) {}`                                                                                                   |
| `new MixedType()`     | `new MixedType(true)` | `Only booleans are allowed in an if condition, (int\|string) given.` reported for `if ($key) {}`<br><br>`Only booleans are allowed in an if condition, mixed given.` reported for `if ($value) {}` |

Conclusions/assumptions:

- Explicit mixed on `TKey` has no effect, simply because it looks like PHPStan narrows its type to `int|string` regardless (unless explicit types are provided).
- Explicit mixed on `TValue` can have side-effects and works as I expected, or my test is wrong.

So, maybe, if we are not exactly sure whether it is mixed or not, we should not mark it as explicit. Because we can't guarantee what type it actually is, so why do we block such conditions? Maybe it is specified in PHPDoc or something like that. I read `TRUE` like this: I'm certain that this type contains mixed types of data that can't be used as a boolean condition.

For example, with `false/true`, this condition will not report an error `if ($value instanceof \Stringable) {}`, because we do not simply rely on the variable type but are narrowing it with `instanceof`.

Maybe it is for cases, when you want to force the user of extension to clearly specify types for such data. Kind of a strict mode or so.

Not sure what to do, but it sounds like remove `true` is much safer?

PR #858, opened 2025-04-11
2.x: Fix assertAll causes staticMethod.alreadyNarrowedType

### LINKED ISSUE #857: assertAll causes staticMethod.alreadyNarrowedType error
# Bug report

Another follow up to https://github.com/mglaman/phpstan-drupal/issues/852

`assertAllArrays` provides a narrowed type of `array<array>`. Calllng `Inspector::assertAll` returns the same narrowed type, which I think causes the rule error.

Since `assertAll` is more generic, we should make sure it returns the original type if the type is more specific or the same.

[](https://git.drupalcode.org/project/project_browser/-/blob/2.0.x/src/ProjectBrowser/Project.php?ref_type=heads#L77)

Niklan:
Closes #857

### DIFF
--- a/src/Type/InspectorTypeExtension.php
+++ b/src/Type/InspectorTypeExtension.php
@@ -107,6 +107,16 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
             return new SpecifiedTypes();
         }
 
+        $traversable = $node->getArgs()[1]->value;
+        $traversableInfo = $scope->getType($traversable);
+
+        // If it is already not mixed (narrowed by other code, like
+        // '::assertAllArray()'), we could not provide any additional
+        // information. We can only narrow this method to 'array<mixed, mixed>'.
+        if (!$traversableInfo instanceof MixedType) {
+            return new SpecifiedTypes();
+        }
+
         return $this->typeSpecifier->create(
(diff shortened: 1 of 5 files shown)


src/Type/InspectorTypeExtension.php
```diff
@@ -107,6 +107,16 @@ private function specifyAssertAll(MethodReflection $staticMethodReflection, Stat
             return new SpecifiedTypes();
         }
 
+        $traversable = $node->getArgs()[1]->value;
+        $traversableInfo = $scope->getType($traversable);
+
+        // If it is already not mixed (narrowed by other code, like
+        // '::assertAllArray()'), we could not provide any additional
+        // information. We can only narrow this method to 'array<mixed, mixed>'.
+        if (!$traversableInfo instanceof MixedType) {
```
mglaman, line 116:
```suggestion
        if (!$traversableInfo->equals(new MixedType())) {
```

Can we use `equals` to avoid calls of `instanceof`? Lots of code in phpstan-drupal needs to be fixed, but PHPStan has slowly been moving away from performing these kind of checks on types over using its type methods

Niklan: Yes, it works the same :+1:

PR #861, opened 2025-04-14
1.x: Fix negation types for Inspector

Niklan:
Port of #860

### DIFF
--- a/src/Type/InspectorTypeExtension.php
+++ b/src/Type/InspectorTypeExtension.php
@@ -110,8 +110,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
         return $this->typeSpecifier->create(
             $node->getArgs()[1]->value,
             new IterableType(new MixedType(), new MixedType()),
-            TypeSpecifierContext::createTruthy(),
-            false,
+            $context,
             $scope,
         );
     }
@@ -124,8 +123,7 @@ final class InspectorTypeExtension implements StaticMethodTypeSpecifyingExtensio
         return $this->typeSpecifier->create(
             $node->getArgs()[0]->value,
             new IterableType(new MixedType(), new StringType()),
-            TypeSpecifierContext::createTruthy(),
-            false,
+            $context,
             $scope,
         );
     }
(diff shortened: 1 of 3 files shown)


mglaman (approved):

tests/src/Type/data/inspector.php
```diff
@@ -12,82 +12,146 @@ function mixed_function(): mixed {
...
 $input = mixed_function();
 $callable = is_string(...);
-assert(Inspector::assertAll($callable, $input));
-assertType('iterable', $input);
+if (Inspector::assertAll($callable, $input)) {
+    assertType('iterable', $input);
+}
+else {
+    assertType('mixed~iterable', $input);
+}
```
Niklan, line 29: @mglaman it's kind of works, but I still don't like it.

```php
$input = mixed_function();
$callable = is_string(...);
if (Inspector::assertAll($callable, $input)) {
    assertType('iterable', $input);
} else {
    assertType('mixed~iterable', $input);
}
```

Actually, it doesn't mean that `$input` is not `iterable` after that `::assertAll()`. It just means it does not contain strings, but it can still be iterable... I think in `else`, it should be `mixed`, as we can't guarantee it is not iterable. We can't narrow it down to `mixed~iterable<string>`, so we better fallback to `mixed`.

mglaman: I see what you mean, but it is never intended to be used in conditionals like this – it's meant to be used in `asserts` which fail if `false`. So I think it's OK that the condition fails. Because none of the type specifying `assert` extensions would work

src/Type/InspectorTypeExtension.php
```diff
@@ -107,10 +107,23 @@ private function specifyAssertAll(MethodReflection $staticMethodReflection, Stat
...
+        }
+
         return $this->typeSpecifier->create(
             $node->getArgs()[1]->value,
             new IterableType(new MixedType(), new MixedType()),
-            TypeSpecifierContext::createTruthy(),
+            $context,
             false,
             $scope,
         );
```
Niklan, line 129: @mglaman now it works much safer:

```
// Inspector::assertAll()
$callable = fn (string $value): bool => $value === 'foo';
$input = mixed_function();
if (Inspector::assertAll($callable, $input)) {
    assertType("iterable", $input);
}
else {
    assertType('mixed', $input);
}
```

We can only confirm that it is `iterable` if the result is `TRUE`, but when it's `FALSE`, we cannot be sure what the type is, so we should keep its original type, which is `mixed`. 

It might be possible to narrow it based on the callable, but I think for now it is too complex and not worth the effort. We can always improve it later.

mglaman: This works for me, and good call on the comment

Issue #635, opened 2023-11-11
If Drupal cannot be detected, fallback to using Composer to detect the installed path


mglaman:
# Feature request

Various hacks are needed if `drupal/core` is in the vendor directory. The `drupal-finder` package doesn't support this (and probably shouldn't.) But we can in the autoloader: https://github.com/mglaman/phpstan-drupal/blob/main/src/Drupal/DrupalAutoloader.php#L60-L69

```php
        $drupalRoot = $finder->getDrupalRoot();
        $drupalVendorRoot = $finder->getVendorDir();
        if (! (bool) $drupalRoot || ! (bool) $drupalVendorRoot) {
            throw new \RuntimeException("Unable to detect Drupal at {$drupalParams['drupal_root']}");
        }
```

Instead of throwing a runtime exception here, try using Composer's installed paths to find `drupal/core`:

```php
\Composer\InstalledVersions::getInstallPath('drupal/core')
```

Then less hacks are needed.

mglaman:
See https://github.com/webflo/drupal-finder/pull/64

AlexSkrypnyk:
I've added a PR to provide a temp workaround
https://github.com/mglaman/phpstan-drupal/pull/702

PR #873, opened 2025-05-03
[#635] Added support for env variables for Drupal root.

AlexSkrypnyk:
#635

Update of https://github.com/mglaman/phpstan-drupal/pull/702

This helps in cases when PHPStan is called outside of the main website location.

This is still relevant in cases when tooling is outside of Drupal install.

Feel free to close, but it may help someone else.


mglaman:
I'm curious what is the use case? There is no `require-dev` in the root repo? Or for allowing dev dependencies not blocked by Drupal's own requirement?

PR #891, opened 2025-07-08
TimestamItem value is int

### LINKED ISSUE #890: Data type of TimestamItem value should be int not string
https://github.com/mglaman/phpstan-drupal/blob/42cca54ee8bccec83a674ac45d1d17586777187e/stubs/Drupal/Core/Field/Plugin/Field/FieldType/TimestampItem.stub#L8

* Data definition of `value` property in `TimestampItem` is [timestamp](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Field/Plugin/Field/FieldType/TimestampItem.php#L43)
* Parent class of `Timestamp` data type plugin is [IntegerData](https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/TypedData/Plugin/DataType/Timestamp.php#L17)

znerol:
Fixes #890

### DIFF
--- a/stubs/Drupal/Core/Field/Plugin/Field/FieldType/TimestampItem.stub
+++ b/stubs/Drupal/Core/Field/Plugin/Field/FieldType/TimestampItem.stub
@@ -5,7 +5,7 @@ namespace Drupal\Core\Field\Plugin\Field\FieldType;
 use Drupal\Core\Field\FieldItemBase;
 
 /**
- * @property string $value
+ * @property int $value
  */
 class TimestampItem extends FieldItemBase {
 }


znerol:
Will fix the tests when somebody has validated the issues.

mglaman (commented):
I know it's _supposed_ to be an int, but does it actually get casted when using Field API? `\Drupal\Core\TypedData\Plugin\DataType\Timestamp` does extend `IntegerData`, but Field API calls `getValue` not `getCastedValue`. So if loaded from the DB the value is `string`.

znerol:
Wouldn't that apply to [IntegerItem](https://github.com/mglaman/phpstan-drupal/blob/main/stubs/Drupal/Core/Field/Plugin/Field/FieldType/IntegerItem.stub#L6) as well?

PR #899, opened 2025-08-04
Allow installers v2 so we can get rid of PHP 8.4 errors

dpi:
>  Implicitly marking parameter $filesystem as nullable is deprecated
> Composer/Installers/Installer.php:136

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     },
     "require-dev": {
         "behat/mink": "^1.10",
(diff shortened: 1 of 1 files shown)


dpi:
Not sure if you just want to update to v2 wholesale, do what you feel is right.

mglaman (approved):

mglaman:
We'll support both, seems easy enough

PR #900, opened 2025-08-04
Narrow return value of `EntityInterface->id()` by previous calls to `isNew` on an instance of `EntityInterface`

dpi:


mglaman (approved):

mglaman:
I like it. Too bad we can't pull it off with just stubs

PR #912, opened 2025-08-27
Add rule to detect redundant trait usage

AronNovak:
This PR adds a new PHPStan rule `NoRedundantTraitUseRule` that detects when a class uses traits redundantly - specifically when a class uses both a trait A and trait B, where trait A already uses trait B (either directly or transitively).

## Motivation

This rule was initially developed for a Drupal project where it significantly improved PHPStan analysis performance by identifying and eliminating redundant trait usage.


AronNovak:
@mglaman Thanks for the suggestion and the swift response.
https://github.com/phpstan/phpstan-strict-rules/pull/279 - then we aim higher :smiley_cat:

PR #946, opened 2026-03-05
Add stubs for formattable arguments

### LINKED ISSUE #1019: Markup context should not be nullable
Ha team phpstan-drupal!

Today we got the following error:

```
TypeError: Drupal\\Component\\Utility\\Html::escape(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/app/core/lib/Drupal/Component/Render/FormattableMarkup.php on line 238 in Drupal\\Component\\Utility\\Html::escape() (line 433 of /var/www/html/app/core/lib/Drupal/Component/Utility/Html.php
````
And a custom module was the cause, FormattableMarkup (and TranslatableMarkup) allows `null` in the `$arguments` array.
This should be illigal, since they put the value (straight) trough Html::escape, where the argument is typed to a non nullable `string`

tldr:
places that call `Html::escape` should be checked to be string/stringable
can we type-hint FormattableMarkup (and TranslatableMarkup) so we can staticly analyze these places

### PR DESCRIPTION
**Pull request by Niklan**
This is just a proposal, but it could be helpful.

From time to time I'm facing a very straightforward error:

```
php.ERROR: TypeError: Drupal\Component\Utility\Html::escape(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/web/core/lib/Drupal/Component/Render/FormattableMarkup.php on line 238 in Drupal\Component\Utility\Html::escape() (line 433 of /var/www/html/web/core/lib/Drupal/Component/Utility/Html.php). {"severity_level":3,"exception":"[object] (TypeError(code: 0): Drupal\\Component\\Utility\\Html::escape(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/web/core/lib/Drupal/Component/Render/FormattableMarkup.php on line 238 at /var/www/html/web/core/lib/Drupal/Component/Utility/Html.php:433)"}
```

This happens because method `\Drupal\Component\Utility\Html::escape` has strict `string` type for its argument:

```php
public static function escape(string $text): string {
  return htmlspecialchars($text, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
}
```

This was done in that issue [#3441394](https://www.drupal.org/project/drupal/issues/3441394) (basically Drupal 11.0+).

The problem is that `FormattableMarkup`, `TranslatableMarkup`, `PluralTranslatableMarkup` and `TranslationInterface` have very weak PHPDoc type-hints in core for the `$arguments` parameter, which is just `@param array $arguments`.

It means, a simple `(string) new TranslatableMarkup('@name is cool', ['@name' => NULL])` will end up with an error above. The other problem, which is not related to `Html::escape`, but still, very heavily related, is that the key for `$arguments` array can't be an empty string, because in `\Drupal\Component\Render\FormattableMarkup::placeholderFormat` we have:

```php
foreach ($args as $key => $value) {
  switch ($key[0]) {
    // Other switched for '%', '!', ':'.
    default:
      if (!ctype_alnum($key[0])) {
        // Warn for random placeholders that won't be replaced.
        trigger_error(sprintf('Placeholders must begin with one of the following "@", ":" or "%%", invalid placeholder (%s) with string: "%s"', $key, $string), E_USER_WARNING);
      }
      // No replacement possible therefore we can discard the argument.
      unset($args[$key]);
      break;
  }
}
```

It means that an empty string is essentially not a valid value for the key either.

This PR adds stubs that narrow the `$arguments` parameter for those classes and the interface to: `@param array<non-empty-string, string|\Stringable> $args`

I think these stubs can be very helpful, because any such detected problem is a fatal error on production anyway. Sometimes they are very tricky to trace down, and most of the problems come from the weak default PHPDoc types that PHPStan can't simply detect during analysis.

P.S. I'm not entirely sure about `non-empty-string`, because while it is clearly needed here, the problem is that if you're passing data with just `string`, PHPStan will start complaining that the string shouldn't be empty. So you'd have to update upper-level code to make sure the string is not actually empty, by using asserts before the call or updating your PHPDoc types that return values for it. But keys are usually entered manually, so it shouldn't be a huge issue.

Fixes #1019

### DIFF
--- a/stubs/Drupal/Component/Render/FormattableMarkup.stub
+++ b/stubs/Drupal/Component/Render/FormattableMarkup.stub
@@ -1,0 +1,14 @@
+<?php
+
+namespace Drupal\Component\Render;
+
+class FormattableMarkup implements MarkupInterface {
+
+  /**
+   * @param string $string
+   * @param array<non-empty-string, string|\Stringable> $arguments
+   */
+  public function __construct($string, array $arguments) {
+  }
+
+}
--- a/stubs/Drupal/Core/StringTranslation/PluralTranslatableMarkup.stub
+++ b/stubs/Drupal/Core/StringTranslation/PluralTranslatableMarkup.stub
@@ -1,0 +1,18 @@
+<?php
+
+namespace Drupal\Core\StringTranslation;
+
+class PluralTranslatableMarkup extends TranslatableMarkup {
+
+  /**
+   * @param int $count
+   * @param string $singular
+   * @param string $plural
+   * @param array<non-empty-string, string|\Stringable> $args
+   * @param array<string, mixed> $options
+   * @param TranslationInterface|null $string_translation
+   */
+  public function __construct($count, $singular, $plural, array $args = [], array $options = [], $string_translation = NULL) {
+  }
+
+}
--- a/stubs/Drupal/Core/StringTranslation/TranslatableMarkup.stub
+++ b/stubs/Drupal/Core/StringTranslation/TranslatableMarkup.stub
@@ -1,0 +1,18 @@
+<?php
+
+namespace Drupal\Core\StringTranslation;
+
+use Drupal\Component\Render\FormattableMarkup;
+
+class TranslatableMarkup extends FormattableMarkup {
+
+  /**
+   * @param string $string
+   * @param array<non-empty-string, string|\Stringable> $arguments
+   * @param array<string, mixed> $options
+   * @param TranslationInterface|null $string_translation
+   */
+  public function __construct($string, array $arguments = [], array $options = [], $string_translation = NULL) {
+  }
+
+}
--- a/stubs/Drupal/Core/StringTranslation/TranslationInterface.stub
+++ b/stubs/Drupal/Core/StringTranslation/TranslationInterface.stub
@@ -1,0 +1,31 @@
+<?php
+
+namespace Drupal\Core\StringTranslation;
+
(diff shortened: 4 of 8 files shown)


Niklan:
@mglaman it is kind of expected to fail on Drupal 10.4 — it doesn't have a strict type for `Html::escape`, so I'll simply ignore it for now. Both tests can be skipped or updated based on your decision.

All that stuff can also be converted into Rules, so it will give more control over that, but the stubs here look like a simpler solution.

mglaman:
I pushed changes to consolidate this with #1020, which stubs the same classes for #1019.

**Placeholder values are `scalar|\Stringable`, keys are `string`.** `FormattableMarkup` does not declare `strict_types`, so `'@count' => $count` is coerced and works. With `string|\Stringable`, every int placeholder was an `argument.type` error. `non-empty-string` keys also flagged `[$key => $value]` for any plain `string $key`. Only `null` and arrays break at runtime now:

```php
new FormattableMarkup('@count', ['@count' => $count]); // ok
new FormattableMarkup('@name', ['@name' => $name]);    // error when $name is ?string
```

**`t()` and `formatPlural()` are covered.** Most code goes through `StringTranslationTrait` or the global `t()`, so I added stubs for both along with fixture cases. `$this->t('@x', ['@x' => null])` is now reported.

I also merged `main` and fixed the `InstantiationRule` construction for PHPStan 2.2. Your core issue (#3577295) is still the right long-term fix. These stubs cover it until then.

Issue #825, opened 2025-02-11
Call to method Drupal\Core\Entity\Query\QueryInterface::accessCheck() will always evaluate to true (PHPStan 2)


WalkingDexter:
# Bug report

After upgrading to PHPStan 2, errors related to the `accessCheck()` method occurred. A similar problem was previously solved in #508.

Versions used:

- phpstan/phpstan `2.1.4`
- mglaman/phpstan-drupal `2.0.1`

### Code snippet that reproduces the problem

```php
// Error: method.alreadyNarrowedType
\Drupal::entityQuery('node')
  ->accessCheck()
  ->execute();

// Error: method.alreadyNarrowedType
\Drupal::entityQuery('node')
  ->accessCheck()
  ->condition('type', 'article')
  ->execute();

// No error.
\Drupal::entityQuery('node')
  ->condition('type', 'article')
  ->accessCheck()
  ->execute();

// Error: method.alreadyNarrowedType
$query = \Drupal::entityQuery('node');
$query->condition('type', 'article');
$query->accessCheck();
$query->execute();
```

Public example from GitLab CI: https://git.drupalcode.org/project/simple_sitemap/-/jobs/4303395

Niklan:
Having the same issue in some projects, I can't replicate it with just testing:

```php
<?php

use function PHPStan\Testing\assertType;

// Error: method.alreadyNarrowedType
$result = \Drupal::entityQuery('node')->accessCheck()->execute();
assertType('array<int, string>', $result);

// Error: method.alreadyNarrowedType
$result = \Drupal::entityQuery('node')->accessCheck()->condition('type', 'article')->execute();
assertType('array<int, string>', $result);

// No error.
$result = \Drupal::entityQuery('node')->condition('type', 'article')->accessCheck()->execute();
assertType('array<int, string>', $result);

// Error: method.alreadyNarrowedType
$query = \Drupal::entityQuery('node');
$query->condition('type', 'article');
$query->accessCheck();
$result = $query->execute();
assertType('array<int, string>', $result);
```

```php
wodby@php.container:/var/www/html $ phpunit --filter EntityQueryDynamicReturnTypeExtensionTest
PHPUnit 9.6.22 by Sebastian Bergmann and contributors.

Warning:       No code coverage driver available

................................                                  32 / 32 (100%)

Time: 00:00.510, Memory: 150.50 MB

OK (32 tests, 32 assertions)
```

* PHPStan 2.1.5
[...]

PR #951, opened 2026-03-16
Fix method.alreadyNarrowedType from all QueryInterface::accessCheck()s (#825)

travis-bradbury:
This includes a test with the examples from https://github.com/mglaman/phpstan-drupal/issues/825#issuecomment-2661420792.


mglaman (approved):
This is what was needed!!? Thank you

Issue #929, opened 2026-01-11
Define entities and fields as universalObjectCratesClasses?


mfb:
# Bug report

Entities and fields have magic getters and setters allowing them to store basically any property.  This is used in core, for example adding a `_referringItem` property to an entity when it's referenced, and adding multiple database columns to a field.

So arguably, entity and field base classes/interfaces should be added to a universalObjectCratesClasses list? These would be `Drupal\Core\Entity\EntityBase`, `Drupal\Core\Entity\EntityInterface` (useful since variables/arguments are often typed to an interface), `Drupal\Core\Field\FieldItemBase`, `Drupal\Core\Field\FieldItemInterface` 

### Code snippet that reproduces the problem

Didn't make a code snippet yet but I could.

mglaman:
Picking this up. #1048 was filed for the same symptom without noticing this issue, and #1053 implements the narrowest slice: `Drupal\Core\Entity\ContentEntityInterface` is added to `universalObjectCratesClasses`. That opens PHPStan's dynamic-property gate for the interface and everything extending it, so `$entity->field_foo` works on a value typed `ContentEntityInterface` or `NodeInterface`, and `@property` tags declared on those interfaces start resolving too.

On the other entries proposed here, checked against core source:

- **`EntityInterface`**: defensible on 11.2 and later. `EntityBase` implements `__get()` and `__set()` for any name there, and `ConfigEntityBase` carries `#[AllowDynamicProperties]`, so config entities are crates at runtime. On 10.4 the base class has no magic methods, so listing the interface would claim more than the runtime does on that version.
- **`EntityBase` and `FieldItemBase`**: redundant. Both declare `__get()`, so PHPStan already allows dynamic properties on them and their subclasses.
- **`FieldItemInterface`**: also redundant for the gate. Core's interface declares `__get()`, `__set()`, and `__isset()` itself. Whether `$item->anything` resolves is a question of which reflection extension answers, not of the crate list.

Proposal: this issue tracks the `EntityInterface` decision and the `_referringItem` case from the description as a fixture. #1053 stays scoped to `ContentEntityInterface`, where the runtime behavior is the same on every supported core version.

PR #958, opened 2026-04-03
Allow all magic properties on FieldItemListInterface

### LINKED ISSUE #782: Add @mixin to FieldItemListInterface to explain that item properties are being forwarded
# Feature request

Using `@mixin` it is possible to explain that instances of FieldItemListInterface forward the properties its items:

```php
/**
 * @template T of FieldItemInterface
 * @extends ListInterface<T>
 * @property mixed $value
 * @mixin T
 */
interface FieldItemListInterface extends ListInterface {
}
```

This would allow for things like:
```php
/** @var \Drupal\Core\Field\FieldItemListInterface<\Drupal\text\Plugin\Field\FieldType\TextWithSummaryItem> $node_body */
$node_body = $node->get('body');

$body = $node_body->processed;
```

Unfortunately `@mixin` also forwards methods. Hopefully `@mixin-properties` will be introduced, or  `@mixin` will auto-detect what is being forwarded: 

[phpstan/phpstan #11259: Fine-grained mixin support](https://github.com/phpstan/phpstan/issues/11259)

marklabrecque-ab:
## Summary

Fixes #782 and #929.

`FieldItemList::__get()` delegates to the first item's properties, so any property on the underlying `FieldItem` should be valid — not just `entity`, `value`, and `target_id`. Properties like `format`, `processed`, `summary`, `alt`, `title`, `uri`, etc. are all commonly used but previously triggered PHPStan errors due to the hardcoded allowlist.

### Changes

- **`FieldItemListPropertyReflection::canHandleProperty()`** now returns `true` for any property name, removing the hardcoded allowlist. This matches the pattern already used for `ContentEntityInterface` in `EntityFieldsViaMagicReflectionExtension::hasProperty()`. The existing `@todo` comment in the code already suggested this change.
- **Fallback type** changed from `NullType` to `MixedType` in `getReadableType()` and `getWritableType()`. For properties not in the known set (`entity`, `value`, `target_id`), `MixedType` is more accurate than `NullType` since delegated properties could be any type.
- Updated tests to reflect the new behavior.


mglaman:
Looks like these are expected now

Issue #586, opened 2023-07-03
Call to deprecated method loadRevision() of class Drupal\Core\Entity\EntityStorageInterface


driskell:
# Bug report

```
  x    Call to deprecated method loadRevision() of class Drupal\Core\Entity\EntityStorageInterface:
         in drupal:10.1.0 and is removed from drupal:11.0.0. Use
         \Drupal\Core\Entity\RevisionableStorageInterface::loadRevision instead.
```

Using phpstan-drupal 1.1.36

### Code snippet that reproduces the problem

```
/** @var \Drupal\Core\Entity\ContentEntityStorageInterface $storage */
$storage->loadRevision(1);
```

I think this is fixed previously already in #576 but only for `RevisionableStorageInterface`. I think the issue lies in the way that ContentEntityStorageInterface inherits its interfaces separately with `EntityStorageInterface` and `TranslatableRevisionalStorageInterface` which then inherits `RevisionableStorageInterface` which then inherits `EntityStorageInterface`. I think perhaps that additional `EntityStorageInterface` on the `ContentEntityStorageInterface` gets in the way

mglaman:
Are you able to add stubs to your project to test out if additional stubs are needed?

This is enough of a data sample to add a proper test here. I thought I did, maybe I was lazy and didn't

driskell:
I tried to add stub for ContentEntityStorageInterface but to no avail. I think maybe this might extend into issues with `@not-deprecated` where there is multiple inheritance paths because `ContentEntityStorageInterface` extends from both `EntityStorageInterface` and `TranslatableRevisionableStorageInterface` with the `EntityStorageInterface` specified first. No matter what I did in the stub files (adding `@not-deprecated` for `loadRevision` in all of them for example) I would still get the deprecation error.

heddn:
could we hack core and see if reversing the order of implementing the interfaces fixes things? I'm still seeing this. Will probably just ignore the warnings since it is a false positive, but it would be nice to fix this more properly.

driskell:
> could we hack core and see if reversing the order of implementing the interfaces fixes things? I'm still seeing this. Will probably just ignore the warnings since it is a false positive, but it would be nice to fix this more properly.

I recall trying various combinations including removing inheritance and different inheritance and I THINK also reordering etc but to no avail - wasn’t sure if a cache was kicking in but do remember using debug mode - just ran out of available time to trace it. Worth someone doing some digging

driskell:
[...]

PR #980, opened 2026-04-21
Fix wrongly deprecated method RevisionableStorageInterface::loadRevision

TheFakeCake:
Since `RevisionableStorageInterface` extends `EntityStorageInterface`, the deprecations are also inherited.

The `loadRevision()` method is not supposed to be deprecated on `RevisionableStorageInterface` though.

### DIFF
--- a/stubs/Drupal/Core/Entity/RevisionableStorageInterface.stub
+++ b/stubs/Drupal/Core/Entity/RevisionableStorageInterface.stub
@@ -7,11 +7,13 @@ interface RevisionableStorageInterface extends EntityStorageInterface {
   /**
    * @param int|numeric-string|string $revision_id
    * @return \Drupal\Core\Entity\RevisionableInterface|null
+   * @not-deprecated
    */
   public function loadRevision($revision_id);
 
   /**
    * @param int|numeric-string|string $revision_id
+   * @not-deprecated
    */
   public function deleteRevision($revision_id): void;
   


mglaman:
Did this work for you? We tried it in https://github.com/mglaman/phpstan-drupal/pull/576 and follow up in https://github.com/mglaman/phpstan-drupal/issues/586

TheFakeCake:
In my Drupal 10.6.8 project, the following code gives me a PhpStan error:
```php
$storage = \Drupal::entityTypeManager()->getStorage('paragraph');

if (!$storage instanceof RevisionableStorageInterface) {
  return null;
}

foreach ($node->get('field_content')->getValue() as $content) {
  if (isset($content['target_revision_id'])) {
    $firstLevelParagraph = $storage->loadRevision($content['target_revision_id']); // <----ERROR
    // ...
  }
}
```
```
Call to deprecated method loadRevision() of interface Drupal\Core\Entity\RevisionableStorageInterface:
in drupal:10.1.0 and is removed from drupal:11.0.0. Use
\Drupal\Core\Entity\RevisionableStorageInterface::loadRevision instead.
🪪  method.deprecated
```
This is obviously a false positive and after some research I found https://phpstan.org/writing-php-code/phpdocs-basics#deprecations where the `@not-deprecated` is mentionned.

When I add `@not-deprecated` to the `phpstan-drupal` stub, clear PhpStan cache and analyze again, I don't get the error anymore.

Some details about my project setup:
```
PHP runtime version: 8.1.30
PHP version for analysis: 8.4.19 (from config)

PHPStan version: 2.1.51

Extension installer:
composer/pcre: 3.3.2
mglaman/phpstan-drupal: 2.0.14
phpstan/phpstan-deprecation-rules: 2.0.4
```

I changed my PR commit to add the `@not-deprecated` to `deleteRevision()` too since it has the same false positive problem (I get the same result).

mglaman:
Thanks for the info!

Issue #753, opened 2024-05-02
Determine return value from `\Drupal\Core\Config\Config::get`


mstrelan:
# Feature request

`\Drupal\Core\Config\Config::get` returns mixed.

Given the following code in a functional test:

```
$theme = $this->config('system.theme')->get('default');
$this->assertStringContainsString('foo', $theme);
```

When I analyze it with phpstan level 9 I get:

```
Parameter #2 $haystack of method PHPUnit\Framework\Assert::assertStringContainsString() expects string, mixed given. 
```

However we have config schema for `system.theme` that says that `default` is a string. Perhaps we can figure this out in phpstan-drupal so we don't need to make assertions about the data. However it might not be wise to do this. If we put the same code in to a kernel test then `$theme` is in fact `null`, so phpstan is right to complain about it.

mglaman:
The one hard part I foresee: I'm pretty sure the typed config manager is stateful. If it doesn't rely on the database at all and only plugin discovery maybe we can make this work

simesy:
Why not contribute a method(s) to the class in Drupal that returns explicit types, and even throws an exception if the type is not a match? It seems like methods for scalar types like Config::getString() should be feasible?

mstrelan:
@simesy interesting idea, I just came across this in Symfony's ParameterBag class. 

https://github.com/symfony/http-foundation/blob/7.1/ParameterBag.php#L125-L152

mxr576:
Just a side note, now that constraints are also applicable to config entities, we can go even crazier stuff and parse those as well and tell PHPStan that something is not string, but an enum or such.

Kingdutch:
Is this not exactly what we have the config schema YAML files for?

We should be able to find the config schema based on the name of the config, based on the convention (or rule?) that all config is prefixed with its module name (or by creating an index of all schema files in the repo). That would allow us to generate a dynamically typed `Config<T>` generic stub, where `T` is the property mapping from the schema.yml file.

Both `set` and `get` can then do type validation based on the `T` associated with the generic.

The hardest part is probably doing the `set` type validation against the type being passed in by PHPStan (e.g. array shapes with nested objects?)

Maybe I don't understand Matt's original comment 
> The one hard part I foresee: I'm pretty sure the typed config manager is stateful. If it doesn't rely on the database at all and only plugin discovery maybe we can make this work

What's the stateful part?

mstrelan:
[...]

PR #993, opened 2026-06-15
Config::get return type extension

mstrelan:
I vibe-coded a solution to #753. Please accept my apologies, as I don't have enough experience with this repo to verify if this is a good approach, so please only review as a proof of concept.

The TL;DR is:

- Narrows `Config::get()` return types using Drupal's config schema files, so calls like `\Drupal::config('system.cron')->get('logging')` resolve to `bool|null` instead of mixed. Only applies to configs with the `FullyValidatable` constraint, which guarantees the schema is complete. Types are always nullable since any key can be absent at runtime, but maybe we should assume they are not?
- Also adds an opt-in rule (`configGetUnknownKeyRule`) that reports calls to `->get()` with a key that doesn't exist in the  schema. This catches typos at analysis time rather than runtime.

I've attached a more verbose explanation, but fair warning about the use of the LLM here.

[config-get-return-type-extension.md](https://github.com/user-attachments/files/28938876/config-get-return-type-extension.md)

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -28,6 +28,7 @@ parameters:
 			entityQuery: true
 			entityRepository: true
 			stubFiles: true
+		configGetReturnType: false
 		rules:
 			testClassSuffixNameRule: false
 			dependencySerializationTraitPropertyRule: false
@@ -270,6 +271,7 @@ parametersSchema:
 			entityRepository: boolean()
 			stubFiles: boolean()
 		])
+		configGetReturnType: boolean()
 		rules: structure([
 			testClassSuffixNameRule: boolean()
 			dependencySerializationTraitPropertyRule: boolean()
@@ -336,6 +338,8 @@ services:
 	-
 		class: mglaman\PHPStanDrupal\Type\ConfigGetDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
+		arguments:
+			configGetReturnType: %drupal.configGetReturnType%
 	-
 		class: mglaman\PHPStanDrupal\Type\DrupalClassResolverDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
--- a/extension.neon
+++ b/extension.neon
@@ -40,6 +40,7 @@ parameters:
 			entityStorageDirectInjectionRule: false
 			symfonyYamlParseRule: false
 			entityOperationsCacheabilityRule: false
+			configGetUnknownKeyRule: false
 		entityMapping:
 			aggregator_feed:
 				class: Drupal\aggregator\Entity\Feed
@@ -281,6 +282,7 @@ parametersSchema:
 			entityStorageDirectInjectionRule: boolean()
 			symfonyYamlParseRule: boolean()
 			entityOperationsCacheabilityRule: boolean()
+			configGetUnknownKeyRule: boolean()
 		])
 		entityMapping: arrayOf(anyOf(
 			structure([
@@ -297,6 +299,8 @@ services:
 		class: mglaman\PHPStanDrupal\Drupal\ServiceMap
 	-
 		class: mglaman\PHPStanDrupal\Drupal\ExtensionMap
+	-
+		class: mglaman\PHPStanDrupal\Drupal\ConfigSchemaData
 	-
 		class: mglaman\PHPStanDrupal\Drupal\EntityDataRepository
 		arguments:
@@ -329,6 +333,9 @@ services:
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 		arguments:
 			containerHasAlwaysTrue: %drupal.bleedingEdge.containerHasAlwaysTrue%
+	-
+		class: mglaman\PHPStanDrupal\Type\ConfigGetDynamicReturnTypeExtension
+		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 	-
 		class: mglaman\PHPStanDrupal\Type\DrupalClassResolverDynamicReturnTypeExtension
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
--- a/src/Drupal/ConfigSchemaData.php
+++ b/src/Drupal/ConfigSchemaData.php
@@ -100,8 +100,14 @@ class ConfigSchemaData
             // Recurse into the element definition.
             if (isset($definition['sequence']) && is_array($definition['sequence'])) {
                 $seq = $definition['sequence'];
-                $elementDef = isset($seq[0]) && is_array($seq[0]) ? $seq[0] : (!isset($seq[0]) ? $seq : null);
-                if ($elementDef !== null && is_array($elementDef)) {
+                /** @var array<mixed>|null $elementDef */
+                $elementDef = null;
+                if (isset($seq[0]) && is_array($seq[0])) {
+                    $elementDef = $seq[0];
+                } elseif (!isset($seq[0])) {
+                    $elementDef = $seq;
+                }
+                if ($elementDef !== null) {
                     return $this->keyExistsInDefinition($elementDef, $parts);
                 }
             }
@@ -217,8 +223,14 @@ class ConfigSchemaData
             }
             if (isset($definition['sequence']) && is_array($definition['sequence'])) {
                 $seq = $definition['sequence'];
-                $elementDef = isset($seq[0]) && is_array($seq[0]) ? $seq[0] : (!isset($seq[0]) ? $seq : null);
-                if ($elementDef !== null && is_array($elementDef)) {
+                /** @var array<mixed>|null $elementDef */
+                $elementDef = null;
+                if (isset($seq[0]) && is_array($seq[0])) {
+                    $elementDef = $seq[0];
+                } elseif (!isset($seq[0])) {
+                    $elementDef = $seq;
+                }
+                if ($elementDef !== null) {
                     return $this->resolveKeyType($elementDef, $parts);
                 }
             }
--- a/src/Drupal/ConfigSchemaData.php
+++ b/src/Drupal/ConfigSchemaData.php
@@ -1,0 +1,287 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Drupal;
+
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\BooleanType;
+use PHPStan\Type\FloatType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\MixedType;
+use PHPStan\Type\NullType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+use PHPStan\Type\UnionType;
+use function array_key_exists;
+use function array_shift;
+use function explode;
+use function in_array;
+use function is_array;
+use function is_string;
+
+class ConfigSchemaData
+{
+    /**
+     * Raw schema definitions keyed by schema type name.
+     *
+     * @var array<string, array<string, mixed>>
+     */
+    private static $definitions = [];
+
+    /**
+     * Config names that have FullyValidatable constraint.
+     *
+     * @var array<string, bool>
+     */
+    private static $fullyValidatable = [];
+
+    /**
+     * @param array<string, array<string, mixed>> $definitions
+     * @param array<string, bool> $fullyValidatable
+     */
+    public function setSchema(array $definitions, array $fullyValidatable): void
+    {
+        self::$definitions = $definitions;
+        self::$fullyValidatable = $fullyValidatable;
+    }
+
+    public function isFullyValidatable(string $configName): bool
+    {
+        return self::$fullyValidatable[$configName] ?? false;
+    }
+
+    /**
+     * Returns true if the given key path exists in the schema for a FullyValidatable config.
+     */
+    public function keyExistsInSchema(string $configName, string $key): bool
+    {
+        if (!$this->isFullyValidatable($configName)) {
+            // Can't validate keys for non-FullyValidatable configs.
+            return true;
+        }
+
+        $definition = self::$definitions[$configName] ?? null;
+        if ($definition === null) {
+            return true;
+        }
+
+        $definition = $this->resolveDefinition($definition);
+        $parts = explode('.', $key);
+        return $this->keyExistsInDefinition($definition, $parts);
+    }
+
+    /**
+     * @param array<string, mixed> $definition
+     * @param string[] $parts
+     */
+    private function keyExistsInDefinition(array $definition, array $parts): bool
+    {
+        $definition = $this->resolveDefinition($definition);
+
+        if ($parts === []) {
+            return true;
+        }
+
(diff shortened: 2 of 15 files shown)


mstrelan:
This passes on 11.3 but fails on 10.4, which kind of proves that this is working, since the same config in 10.4 is not `FullyValidatable`. It falls back to `mixed` which is what we expect.

src/Drupal/ConfigSchemaData.php
```diff
@@ -0,0 +1,287 @@
...
+     */
+    private function keyExistsInDefinition(array $definition, array $parts): bool
+    {
+        $definition = $this->resolveDefinition($definition);
+
+        if ($parts === []) {
+            return true;
+        }
+
+        if (!isset($definition['mapping']) || !is_array($definition['mapping'])) {
```
mglaman, line 218: **Blocking — confirmed false positive on sequence children.** This only walks `mapping`, so a path segment that lands on a `sequence` returns `false` and any child key is reported as unknown. A sequence has arbitrary keys, so this is wrong. Reproduced against `system.mail` (FullyValidatable, `interface` is a sequence):

```php
\Drupal::config('system.mail')->get('interface.default');
// => Config key "interface.default" does not exist in the schema for "system.mail".  ← false positive
```

When traversal hits a `sequence`, accept the remaining key path. The same gap exists in `resolveKeyType` below, but there it just yields `mixed` (safe), so only the rule emits a bad error.

src/Drupal/ConfigSchemaData.php
```diff
@@ -0,0 +1,287 @@
...
+use function is_string;
+
+class ConfigSchemaData
+{
+    /**
+     * Raw schema definitions keyed by schema type name.
+     *
+     * @var array<string, array<string, mixed>>
+     */
+    private static $definitions = [];
```
mglaman: These are `static` on a service that's registered as a singleton — no reason for static state here. It leaks across container instances and risks test cross-contamination (one test's schema bleeding into another). Make `$definitions` and `$fullyValidatable` instance properties.

extension.neon
```diff
@@ -329,6 +333,9 @@ services:
 		tags: [phpstan.broker.dynamicMethodReturnTypeExtension]
 		arguments:
 			containerHasAlwaysTrue: %drupal.bleedingEdge.containerHasAlwaysTrue%
+	-
+		class: mglaman\PHPStanDrupal\Type\ConfigGetDynamicReturnTypeExtension
```
mglaman, line 339: **Blocking — on by default.** Unlike the rule (opt-in via `configGetUnknownKeyRule: false`), this type extension is registered unconditionally, so every user gets `mixed → bool|null` etc. on upgrade. That can surface *new* errors (e.g. a now-`string|null` value passed into a non-null param). Per the versioning policy this can't ship in a patch — either gate it behind a parameter too, or release as a minor and call it out in the changelog.

src/Rules/Drupal/ConfigGetUnknownKeyRule.php
```diff
@@ -0,0 +1,153 @@
...
+                )
+                    ->identifier('drupal.configGetUnknownKey')
+                    ->build();
+            }
+        }
+
+        return $errors;
+    }
+
+    private function resolveConfigName(MethodCall $methodCall, Scope $scope): ?string
```
mglaman: `resolveConfigName()` and `extractFirstStringArg()` are copy-pasted verbatim from `ConfigGetDynamicReturnTypeExtension` (line 79 there) — ~60 lines, and the comments are already drifting apart. Hoist to one place (a trait, or methods on `ConfigSchemaData`) so the four config-name patterns stay in sync.

src/Drupal/DrupalAutoloader.php
```diff
@@ -266,6 +269,93 @@ class: Drupal\jsonapi\Routing\JsonApiParamEnhancer
...
+        foreach ($this->themeData as $extension) {
+            $schemaDir = $this->drupalRoot . '/' . $extension->getPath() . '/config/schema';
+            if (is_dir($schemaDir)) {
+                $schemaDirs[] = $schemaDir;
+            }
+        }
+
+        // Parse all schema files.
+        foreach ($schemaDirs as $dir) {
+            $files = glob($dir . '/*.schema.yml');
```
mglaman: Non-blocking, worth noting: this parses *every* `*.schema.yml` (core + all modules + all themes) on every autoloader boot, even when neither feature is active. On a large site that's real startup latency. Consider lazy-loading the schemas, or only parsing when the extension/rule is enabled.

mglaman (commented):
🤖 review incoming:

Nice proof-of-concept — the core instinct (gate everything on `FullyValidatable` so the schema is authoritative) is the right one, and the conservative `mixed`/`null` fallbacks mean the type extension can only add info, not break existing code. Tests pass locally (14/14).

Left inline notes. Two blocking items: a confirmed sequence false-positive in the rule, and the type extension being on-by-default (release/versioning implications). The rest are smells worth cleaning up before merge. On your open question — yes, keep the type nullable: `Config::get()` returns `null` for an absent key regardless of schema completeness, so `T|null` is correct.

mstrelan:
**1. Sequence traversal fix (BLOCKING) — ConfigSchemaData.php** 

`keyExistsInDefinition()` and `resolveKeyType()` now handle sequence nodes: when a definition has `type: sequence`, the next path segment is treated as a dynamic element key (consumed without validation), then recursion continues into the element definition. This fixes the false positive on `\Drupal::config('system.mail')->get('interface.default')`.

New test cases added:
- Rule fixture: `interface.default` on `system.mail` no longer reports an error
- Type fixture: `assertType('string|null', ...)` for `interface.default`

**2. Opt-in gate for type extension (BLOCKING) — extension.neon + ConfigGetDynamicReturnTypeExtension.php**

Added `drupal.configGetReturnType: false` parameter. The extension returns `null` early when the flag is `false`. A new test config neon (`phpunit-drupal-phpstan-config-get-return-type.neon`) enables it for the type extension tests.

**3. Static property improvements — ConfigSchemaData.php**

Properties kept static (required for PHPStan test infrastructure cross-container survival) but given proper type declarations: `private static array $definitions = []` and `private static array $fullyValidatable = []`, with doc comments explaining why static is necessary.

**4. ConfigNameResolverTrait — new src/Drupal/ConfigNameResolverTrait.php**

`resolveConfigName()` and `extractFirstStringArg()` extracted from both `ConfigGetUnknownKeyRule` and `ConfigGetDynamicReturnTypeExtension` into a shared trait. Both classes now use `ConfigNameResolverTrait`.

**5. Lazy-loading TODO — DrupalAutoloader.php**

Added `// TODO: lazy-load schemas when neither configGetReturnType nor configGetUnknownKeyRule is active.` comment at the top of `loadConfigSchemas()`.

mglaman (changes requested):
Thanks for taking this on — #753 has been wanted for a while, and gating on `FullyValidatable` is the right call. I checked out the branch, ran the suite, and probed the new code against the Drupal 11.3.8 fixture. The new tests pass and phpcs is clean. The one full-suite failure (`InspectorTypeExtensionTest`) is pre-existing on `main` and unrelated.

There are three blockers and a few design notes.

## Blocker 1: memory regression from eager schema loading

`php vendor/bin/phpstan analyze` on this repo crashes at the default 128M limit on this branch. It passes on `main`. `DrupalAutoloader::loadConfigSchemas()` parses every `*.schema.yml` in core plus all modules and themes at bootstrap, unconditionally — your TODO already flags it. The autoloader has the container, so gate on the parameters:

```php
$parameters = $container->getParameter('drupal');
if (!($parameters['configGetReturnType'] ?? false)
    && !($parameters['rules']['configGetUnknownKeyRule'] ?? false)) {
    return;
}
```

## Blocker 2: dynamic type references produce false positives

`resolveDefinition()` cannot resolve `[%parent...]`/`[%key]` references, so the definition ends up without a `mapping` and `keyExistsInDefinition()` returns `false`. That flags valid keys on real core config. `system.mail` is `FullyValidatable`, and `mailer_dsn.options` has `type: mailer_dsn.options.[%parent.scheme]`:

```php
// Valid real-world key, but the rule reports:
// Config key "mailer_dsn.options.verify_peer" does not exist in the schema for "system.mail".
\Drupal::config('system.mail')->get('mailer_dsn.options.verify_peer');
```

When a definition's `type` contains `[`, treat the subtree as unverifiable — return `true` from `keyExistsInDefinition()` and `null` from `resolveKeyType()`, same as the sequence fallback. The rule's pitch is "any key not listed is a genuine error," so false positives here undermine the whole feature.

## Blocker 3: config name resolution never verifies the receiver

`ConfigNameResolverTrait` matches on method name alone. Pattern 1 accepts any static call named `config`, and pattern 2 accepts any instance method named `config`. Both produce false positives:

```php
class NotDrupal {
    public static function config(string $name): Config {
        return \Drupal::config('some.other_config');
    }
}
// Reported: Config key "legit_key_of_other_config" does not exist in the schema for "system.maintenance".
NotDrupal::config('system.maintenance')->get('legit_key_of_other_config');

class HasConfigHelper {
    public function config(string $name): Config {
        return \Drupal::config('mymodule.' . $name);
    }
}
// Reported against the wrong config entirely.
$this->config('system.maintenance')->get('anything_at_all');
```

The `$this->config($name)` prefixing helper is a common contrib pattern. Pattern 1 must verify the class resolves to `\Drupal`; pattern 2 must verify the declaring class uses `ConfigFormBaseTrait`.

## Bug: partial unions drop the unresolvable branch

In `ConfigGetDynamicReturnTypeExtension`, when the key type is a union of constant strings and only some resolve, the extension returns the resolved types and silently drops the rest:

```php
$key = $flag ? 'message' : 'unknown_key_xyz';
// Inferred: string|null. Should stay mixed — one branch is unresolvable.
\Drupal::config('system.maintenance')->get($key);
```

If any constant string fails to resolve, return `null` and skip narrowing.

## Design notes (non-blocking)

- **Static properties on `ConfigSchemaData` are load-bearing.** I converted them to instance state matching `ServiceMap` and 12 of 15 tests fail, so the workaround is real. But it is fragile global state. Before merge, the comment should explain *why* the bootstrap does not repopulate test containers, not just assert that it doesn't.
- **Wildcard schemas are unsupported.** `isFullyValidatable()` is an exact-name lookup, so `block.block.*`, `views.view.*` — every config entity — gets zero coverage. Fine for the initial scope, but document it.
- **No narrowing through variables.** `$config = \Drupal::config('x'); $config->get('y')` stays `mixed`. The longer-term fix is a return type extension on `ConfigFactoryInterface::get()` that produces a config-name-carrying type so the name survives assignment. The syntactic chain-matching is acceptable as a first pass.
- **Keep the `|null`.** Keys can be absent before save and the config object may not exist at all. The conservative union is correct.
- `Choice` constraints could later narrow to literal types, and the two new parameters need README coverage.

The structure is right and the fixes are all localized. Happy to merge once the blockers are addressed.

mglaman:
I pushed ca1af2e addressing the review blockers directly, since the fixes were all localized.

One correction to the review first: the 128M crash I attributed to eager schema loading also reproduces on `main` with a cold result cache. My earlier `main` run passed because the cache was warm. So that specific evidence was wrong. The eager parsing still has a real cost — every schema file of every extension parsed in every parallel worker's bootstrap, whether or not the features are enabled — so the lazy-loading change stands, just without the dramatic crash as justification.

What changed:

**Lazy schema loading.** The bootstrap now only collects schema directories; `ConfigSchemaData::ensureLoaded()` parses the files on the first query. This matters beyond memory: PHPStan executes bootstrap files once per process, so gating the *parsing* on parameters read at bootstrap time would read whichever container bootstraps first — wrong under the test infrastructure and any multi-config process. Deferring the parse to query time sidesteps that, because the consumers (`ConfigGetDynamicReturnTypeExtension` via its constructor flag, the rule via `conditionalTags`) already gate per-container. This is also why the static properties exist, and the comment now explains the actual mechanism: bootstrap-once plus multiple containers, not a vague "test infrastructure" hand-wave.

**Receiver verification.** `resolveConfigName()` now requires static `config()` calls to resolve to `\Drupal`, and instance `config()` calls to come from a class using `ConfigFormBaseTrait`. Both false positives from the review have regression fixtures.

**Dynamic type references.** Definitions whose `type` contains `[` (e.g. `mailer_dsn.options.[%parent.scheme]`) are treated as unverifiable: `keyExistsInDefinition()` accepts keys beneath them and `resolveKeyType()` declines to narrow. `system.mail` fixtures cover both paths.

**Partial unions.** If any constant string in the key type fails to resolve, the extension returns `null` instead of a union missing the mixed branch. Also switched to `TypeCombinator::union()` — `new UnionType()` throws on overlapping members, which a `$flag ? 'message' : 'langcode'` key exposed immediately.

One test-infrastructure note: the rule test now registers `tests/fixtures/config/phpunit-drupal-phpstan-config-get-unknown-key.neon` so the rule's parameter is enabled in its container. Without it the test only passed by reading static state leaked from whichever test loaded schemas first.

Full suite passes (the `InspectorTypeExtensionTest` failure is pre-existing on `main`), phpcs and self-analysis are clean.

Remaining from the review, non-blocking: wildcard schema support (`block.block.*` — every config entity currently gets no coverage), `Choice` constraints as literal types, and README coverage for the two parameters. Fine as follow-ups.

src/Drupal/ConfigSchemaData.php
mglaman: Addressed in aabaf0e: sequence definitions consume the dynamic element segment and recurse into the element definition. Covered by the `interface.default` fixture in tests/src/Rules/data/config-get-unknown-key.php.

src/Drupal/ConfigSchemaData.php
mglaman: Keeping the statics, but now with a verified reason: PHPStan executes bootstrap files once per process, while the test infrastructure creates multiple containers afterward — converting to instance properties breaks 12 of 15 tests. ca1af2e documents the actual mechanism on the properties and moves parsing into a lazy `ensureLoaded()`, so the static state is a parse-once cache rather than bootstrap-written data.

extension.neon
mglaman: Addressed in aabaf0e: the extension is gated behind `drupal.configGetReturnType`, default `false`, so upgrades keep `mixed` until users opt in.

src/Rules/Drupal/ConfigGetUnknownKeyRule.php
mglaman: Addressed in aabaf0e: extracted to `ConfigNameResolverTrait`, shared by the rule and the extension. ca1af2e added receiver verification there so both consumers got the fix once.

src/Drupal/DrupalAutoloader.php
mglaman: Addressed in ca1af2e: the bootstrap only collects schema directories now; parsing happens lazily on first query, and both consumers gate their queries, so disabled features never parse.

mglaman (approved):
All review blockers addressed and verified: lazy schema loading, receiver verification, dynamic type reference handling, and the union narrowing fix, each with regression coverage. CI matrix is green across Drupal 10.4 through 11.x-dev for the first time on this branch. Both features are opt-in, excluded from bleedingEdge.neon, and documented in the README.

Issue #412, opened 2022-05-24
Stub ContentEntityInterface for its iterable template


mglaman:
Given a Drupal content entity type:

```
ContentEntityInterface extends \Traversable
ContentEntityBase extends EntityBase implements \IteratorAggregate, ContentEntityInterface
```

It's iterator is for the field values:

```
  /**
   * {@inheritdoc}
   */
  #[\ReturnTypeWillChange]
  public function getIterator() {
    return new \ArrayIterator($this->getFields());
  }
```

Which:

```
  /**
   * Gets an array of all field item lists.
   *
   * @param bool $include_computed
   *   If set to TRUE, computed fields are included. Defaults to TRUE.
   *
   * @return \Drupal\Core\Field\FieldItemListInterface[]
   *   An array of field item lists implementing, keyed by field name.
   */
  public function getFields($include_computed = TRUE);
```

PR #996, opened 2026-06-24
Populate service map when bootstrap is skipped (e.g. under Rector)

### LINKED ISSUE #995: Service map is empty when running under Rector — bootstrap never executes
phpstan-drupal builds the service map in `DrupalAutoloader::register()`, which runs as a `bootstrapFiles` entry (`drupal-autoloader.php`). PHPStan executes those through `CommandHelper::executeBootstrapFile()`. Any tool that builds the PHPStan container directly never calls `CommandHelper`, so the bootstrap is skipped and the map stays empty.

Rector is the case. Its `PHPStanServicesFactory` calls `ContainerFactory::create()` and stops there — no `CommandHelper`. So under `rector process`, `\Drupal::service('renderer')` resolves to `object` instead of `Drupal\Core\Render\Renderer`. The same goes for everything else that leans on the map. Under `phpstan analyse` it's fine: I see 2075 services mapped. Under Rector the bootstrap never fires.

This matters because Rector uses phpstan-drupal for the type info that drives type-aware rules. With an empty map, every `\Drupal::service('...')->someDeprecatedMethod()` is invisible to those rules and silently skipped — no rewrite where `phpstan analyse` would have had the type.

The obvious Rector-side workaround doesn't work:

```php
$rectorConfig->bootstrapFiles([__DIR__ . '/vendor/mglaman/phpstan-drupal/drupal-autoloader.php']);
```
```
[ERROR] The autoloader did not receive the container.
```

Rector requires bootstrap files with no `$container` in scope, and `drupal-autoloader.php` needs it. Larastan's bootstrap is container-free, so the same trick works for Larastan — see rectorphp/rector#8006.

The fix is to not depend on the bootstrap running. Populate the map during container build instead. A guarded, eagerly-instantiated service does it without touching the existing scan:

```php
final class RectorServiceMapInitializer implements DynamicStaticMethodReturnTypeExtension
{
    public function __construct(Container $container, ServiceMap $serviceMap)
    {
        if ($serviceMap->getServices() === []) {
            (new DrupalAutoloader())->register($container);
        }
    }

    // getClass(): 'Drupal'; isStaticMethodSupported(): false; getTypeFromStaticMethodCall(): null
}
```

Register it with `tags: [phpstan.broker.dynamicStaticMethodReturnTypeExtension]` so PHPStan instantiates it when the broker builds. The empty-map guard keeps a normal `phpstan analyse` run a no-op, since the bootstrap already filled the map.

I've run this against Drupal 11 with Rector 2.4 across a full install (webform, photoswipe). `\Drupal::service('renderer')` resolves to `Renderer` and `\Drupal::service('entity_type.manager')` to `EntityTypeManager`, Rector processes both modules without errors, and standalone `phpstan analyse` behaves exactly as before.

I'm not attached to this shape. Lazy population inside `ServiceMap` would be cleaner if you'd prefer it. Happy to open a PR either way.

---

*Claude was used for research and development of this issue and fix.*

bbrala:
Fixes #995.

The service map is built by `DrupalAutoloader::register()`, which runs as a `bootstrapFiles` entry. PHPStan executes those through `CommandHelper::executeBootstrapFile()`. Tools that build the PHPStan container directly never call `CommandHelper`, so the bootstrap is skipped and the map stays empty. Rector is the case — its `PHPStanServicesFactory` calls `ContainerFactory::create()` and stops there — so `\Drupal::service('renderer')` resolves to `object` instead of `Drupal\Core\Render\Renderer`, and type-aware Rector rules silently skip those nodes.

This registers `RectorServiceMapInitializer`, a no-op `DynamicStaticMethodReturnTypeExtension` that PHPStan instantiates eagerly when the broker builds. If the service map is still empty at that point, it runs the autoloader against the live container. The empty-map guard keeps a normal `phpstan analyse` run a no-op, since the bootstrap already populated the map there.

I went with reusing `DrupalAutoloader::register()` wholesale rather than refactoring the scan out of it, to keep the change small. Lazy population inside `ServiceMap` would be cleaner as the real shape if you'd prefer that — happy to rework.

Tested against Drupal 11 with Rector 2.4 across a full install (webform, photoswipe): `\Drupal::service('renderer')` resolves to `Renderer` and `\Drupal::service('entity_type.manager')` to `EntityTypeManager`, Rector processes both modules without errors, and standalone `phpstan analyse` behaves exactly as before.

---

*Claude was used for research and development of this issue and fix.*

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -263,6 +263,9 @@ services:
 	-
 		class: mglaman\PHPStanDrupal\Drupal\ServiceMap
 	-
+		class: mglaman\PHPStanDrupal\Drupal\RectorServiceMapInitializer
+		tags: [phpstan.broker.dynamicStaticMethodReturnTypeExtension]
+	-
 		class: mglaman\PHPStanDrupal\Drupal\ExtensionMap
 	-
 		class: mglaman\PHPStanDrupal\Drupal\EntityDataRepository
--- a/src/Drupal/RectorServiceMapInitializer.php
+++ b/src/Drupal/RectorServiceMapInitializer.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types=1);
+
+namespace mglaman\PHPStanDrupal\Drupal;
+
+use PhpParser\Node\Expr\StaticCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\DependencyInjection\Container;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\DynamicStaticMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+
+/**
+ * Populates the Drupal service map when the bootstrap file is not executed.
+ *
+ * The service map is built by {@see DrupalAutoloader::register()}, which runs as
+ * a `bootstrapFiles` entry (drupal-autoloader.php). PHPStan executes those through
+ * CommandHelper::executeBootstrapFile(). Tools that build the PHPStan container
+ * directly never call CommandHelper, so the bootstrap is skipped and the map stays
+ * empty — most notably Rector, whose PHPStanServicesFactory calls
+ * ContainerFactory::create() and stops there. The result is that \Drupal::service()
+ * resolves to `object` under Rector instead of the concrete service class.
+ *
+ * This service is registered as a no-op dynamic return type extension purely so
+ * PHPStan instantiates it eagerly when the broker is built. If the map is still
+ * empty at that point, it runs the autoloader against the live container. The
+ * empty-map guard keeps a normal PHPStan run a no-op, since the bootstrap already
+ * populated the map there.
+ *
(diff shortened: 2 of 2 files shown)


mglaman:
Closing, since this is fixed upstream. Rector 2.6.0 ([rector-src#8190](https://github.com/rectorphp/rector-src/pull/8190)) passes the PHPStan `$container` to bootstrap files, so `drupal-autoloader.php` populates the service map under Rector. drupal-rector hands off to it as of [#412](https://github.com/palantirnet/drupal-rector/pull/412).

Outside drupal-rector, add the autoloader as a Rector bootstrap file:

```php
$rectorConfig->bootstrapFiles([__DIR__ . '/vendor/mglaman/phpstan-drupal/drupal-autoloader.php']);
```

Thanks for pushing this upstream, bbrala.

PR #999, opened 2026-07-13
Widen drupal-finder constraint and drop stale baseline entries

### LINKED ISSUE #962: Migrate DrupalAutoloader off deprecated drupal-finder v1 APIs
## Problem

`src/Drupal/DrupalAutoloader.php` uses `DrupalFinder\DrupalFinder` and its `locateRoot()` method, which were deprecated in drupal-finder 1.3.0 and **removed in 2.0.0**. Currently `phpstan-baseline.neon` suppresses 5 errors related to this usage:

- `Call to deprecated method locateRoot() of class DrupalFinder\DrupalFinder`
- `Instantiation of deprecated class DrupalFinder\DrupalFinder`
- Two type mismatch errors caused by the old API returning `bool|string`

When the composer constraint eventually allows drupal-finder 2.x, this will cause a hard runtime crash during analysis.

## Proposed resolution

Replace `DrupalFinder\DrupalFinder` with `DrupalFinder\DrupalFinderComposerRuntime`, which accepts a starting path in its constructor and provides the same `getDrupalRoot()` / `getVendorDir()` methods without the deprecated `locateRoot()` call.

```php
// Before
$finder = new DrupalFinder();
$finder->locateRoot(getcwd());
$this->drupalRoot = $finder->getDrupalRoot();

// After
$finder = new DrupalFinderComposerRuntime();
$this->drupalRoot = $finder->getDrupalRoot();
```

After migrating, the 5 baseline suppressions in `phpstan-baseline.neon` for `DrupalAutoloader.php` should be removed.

mglaman:
## Summary
- Widen `webflo/drupal-finder` constraint to `^1.3.1 || ^2.0` — `DrupalAutoloader` already uses `DrupalFinderComposerRuntime`, which is available and non-deprecated in both major versions.
- Remove 5 stale `ignoreErrors` entries in `phpstan-baseline.neon` referencing `DrupalFinder::locateRoot()`/`DrupalFinder` instantiation — dead code paths, the migration to `DrupalFinderComposerRuntime` already happened in #790.

Closes #962

## Test plan
- [x] `php vendor/bin/phpstan analyze` — no errors (confirms the baseline entries were genuinely dead)
- [x] `php vendor/bin/phpunit` — full suite green (one pre-existing unrelated failure in `InspectorTypeExtensionTest`, confirmed present on `main` too)
- [x] `php vendor/bin/phpcs` — clean

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "symfony/finder": "^6.2 || ^7.0 || ^8.0",
         "symfony/yaml": "^6.2 || ^7.0 || ^8.0",
-        "webflo/drupal-finder": "^1.3.1"
+        "webflo/drupal-finder": "^1.3.1 || ^2.0"
     },
     "require-dev": {
         "behat/mink": "^1.10",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,36 +1,3 @@
 parameters:
-	ignoreErrors:
-		-
-			message: """
-				#^Call to deprecated method locateRoot\\(\\) of class DrupalFinder\\\\DrupalFinder\\:
-				Will be removed in v2\\. Future usage should instantiate
-				  a new DrupalFinder object by passing the starting path to its
-				  constructor\\.$#
-			"""
-			count: 1
-			path: src/Drupal/DrupalAutoloader.php
-
-		-
-			message: """
-				#^Instantiation of deprecated class DrupalFinder\\\\DrupalFinder\\:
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -15,7 +15,7 @@
         "phpstan/phpstan-deprecation-rules": "^2.0",
         "symfony/finder": "^6.2 || ^7.0 || ^8.0",
         "symfony/yaml": "^6.2 || ^7.0 || ^8.0",
-        "webflo/drupal-finder": "^1.3.1"
+        "webflo/drupal-finder": "^1.3.1 || ^2.0"
```
mglaman, line 18: Are we still actually compatible with ^1.3, though? I think we have to drop it.

PR #1020, opened 2026-07-22
feat: add stubs for Markup classes for type hinting

### LINKED ISSUE #1019: Markup context should not be nullable

MakerTim:
Fixes #1019

add type hinting so you prevent Type errors on HTML::escape (null is not a string)

### DIFF
--- a/stubs/Drupal/Component/Render/FormattableMarkup.stub
+++ b/stubs/Drupal/Component/Render/FormattableMarkup.stub
@@ -1,0 +1,13 @@
+<?php
+
+namespace Drupal\Component\Render;
+
+class FormattableMarkup implements MarkupInterface, \Countable {
+
+    /**
+     * @param string $string
+     * @param array<string, string|\Stringable> $arguments
+     */
+    public function __construct($string, array $arguments) {}
+
+}
--- a/stubs/Drupal/Core/StringTranslation/TranslatableMarkup.stub
+++ b/stubs/Drupal/Core/StringTranslation/TranslatableMarkup.stub
@@ -1,0 +1,16 @@
+<?php
+
+namespace Drupal\Core\StringTranslation;
+
+use Drupal\Component\Render\FormattableMarkup;
+
+class TranslatableMarkup extends FormattableMarkup {
+
+  /**
+   * @param string $string
+   * @param array<string, string|\Stringable> $arguments
+   * @param array<string, mixed> $options
+   */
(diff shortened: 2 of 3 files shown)


mglaman (changes requested):
Thanks for picking up #1019. A `null` placeholder reaching `Html::escape(string $text)` is a real `TypeError`, and catching it statically is worth doing. Two problems with this approach:

**Int and float placeholders become errors.** `FormattableMarkup` does not declare `strict_types`, so `'@count' => $count` is coerced to a string and works fine. With `array<string, string|\Stringable>` every one of those is now an `argument.type` error, and that pattern is everywhere in Drupal code:

```php
new FormattableMarkup('@x', ['@x' => $count]); // Parameter #2 $arguments ... expects array<string, string|Stringable>, array<string, int> given.
```

The type needs to exclude only what actually breaks: `array<string, scalar|\Stringable>`.

**The common call paths are not covered.** Most code calls `$this->t()`, `formatPlural()`, or `TranslationInterface::translate()`, not `new TranslatableMarkup()`. None of these report `null` with this PR:

```php
$this->t('@x', ['@x' => null]);
$translation->translate('@x', ['@x' => null]);
```

The empty `TranslationInterface.stub` does not add anything either.

#946 stubs the same classes, plus `PluralTranslatableMarkup` and `TranslationInterface`, and has tests. We're consolidating the fix there, widening the type to `scalar|\Stringable` and adding the `StringTranslationTrait` stubs. I'll close this one in favor of #946 once that lands.

mglaman:
Closing in favor of #946, which now covers the same stubs with `scalar|\Stringable` placeholder values, adds `t()` and `formatPlural()`, and closes #1019. Thanks for reporting and kicking this off, MakerTim.

PR #1023, opened 2026-08-04
Do not fatal when a shorthand service class cannot be loaded

mglaman:
## Problem

Service map generation calls `class_exists()` on shorthand service IDs (added in #817). `class_exists()` autoloads the class, and PHP resolves parent classes at declaration time. A service class extending a class from an unavailable module throws an `Error` that kills the bootstrap before PHPStan analyzes a single file:

```
Error thrown in modules/canvas_headless/src/RenderConverter/JsComponentCanvasRenderConverter.php
on line 15 while loading bootstrap file vendor/mglaman/phpstan-drupal/drupal-autoloader.php:
Class "Drupal\custom_elements\RenderConverter\CanvasRenderConverter" not found
```

Canvas hit this: `canvas_headless` registers a decorator for a `custom_elements` service with `decoration_on_invalid: ignore`, and `drupal/custom_elements` sits in `require-dev`. Symfony ignores the decorator at runtime. phpstan-drupal fataled anyway, because the crash happens during class load, not container compilation. See https://git.drupalcode.org/project/canvas/-/merge_requests/1460, which works around it with an explicit `class:` key.

## Fix

Catch `Throwable` around both shorthand `class_exists()` calls (`DrupalAutoloader`, `ServiceMap`) and skip the service when its class cannot load. `ServiceMap::setDrupalServices()` also skips decorator registration when the decorating service was dropped — previously an undefined array key.

## Tests

- New fixture module `service_map_broken` with a shorthand service whose class extends a class from a missing module. Every bootstrap-dependent test now exercises the failure path; without the fix, `DrupalContainerDynamicReturnTypeTest` fatals with the exact error above.
- `ServiceMapFactoryTest::testShorthandServiceWithUnloadableClass` covers `ServiceMap` directly: the broken service is skipped and the decorated service survives with zero decorators.

### DIFF
--- a/src/Drupal/DrupalAutoloader.php
+++ b/src/Drupal/DrupalAutoloader.php
@@ -237,7 +237,7 @@ class DrupalAutoloader
                 //     tags:
                 //       - { name: foo_bar }
                 // @endcode
-                if (!isset($serviceDefinition['class']) && class_exists($serviceId)) {
+                if (!isset($serviceDefinition['class']) && $this->classExists($serviceId)) {
                     $serviceDefinition['class'] = $serviceId;
                 }
                 // @todo sanitize "calls" and "configurator" and "factory"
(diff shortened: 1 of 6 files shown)


mglaman:
Pushed a second commit that respects `decoration_on_invalid: ignore`: when the decorated service does not exist and the decorator opts into `ignore`, the decorating service is dropped from the service map. This matches Symfony's compile-time behavior — the service does not exist in the runtime container either. `null` and the default `exception` keep the current behavior.

longwave:
+1, running into exactly this with PHPStan on a Canvas project that does not use `canvas_headless`.

PR #1032, opened 2026-08-05
Deny uses of AccessResult in boolean context.

travis-bradbury:
This denies use of AccessResultInterface in boolean context: `if ($result)` is always true. It must be `if ($result->isAllowed()` or the like. Further examples are in tests/src/Rules/data/access-result-boolean-context.php.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -34,6 +34,7 @@ parameters:
 			testClassSuffixNameRule: true
 			dependencySerializationTraitPropertyRule: true
 			accessResultConditionRule: true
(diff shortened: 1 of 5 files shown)


mglaman:
Doesn't PHPStan already catch this? If evaluating a non-bool in logic?

travis-bradbury:
Yes, there's a bunch of *.alwaysFalse and *.alwaysTrue rules that catch this. Apparently my issue was an old bug ignored in my baseline 🤦‍♂️

PR #1041, opened 2026-09-03
EntityRepositoryReturnTypeExtension: Handle non-constant strings (fixes #1040)

### LINKED ISSUE #1040: EntityRepositoryInterface::getActive returns *NEVER* when the type argument is just string.
# Bug report

When calling `EntityRepositoryInterface::getActive($entity_type_id, $entity_id, ?array $contexts = NULL);` method with a non-constant string, then it unexpectedly infers as Never.

A very minimal example is this:

```php
<?php

use \Drupal\Core\Entity\EntityRepositoryInterface;
use \Drupal\Core\Entity\EntityInterface;

use function \assert;

function something(EntityRepositoryInterface $repo, EntityInterface $entity): void{
  $entity_id = $entity->id();
  assert($entity_id !== NULL);
  $result = $repo->getActive($entity->getEntityTypeId(), $entity_id);
  \PHPStan\dumpType($result);
}
```
This also reproduces [on the playground](https://phpstan-drupal.mglaman.dev/r/e5c7d1fb-d85b-4bcc-b8d6-3c8bed4205d6). 

For context, this came up when writing the equivalent of a `Drupal\node\Controller\NodeController` for a custom entity type.
That more realistic use case can also be found [on the playground](https://phpstan-drupal.mglaman.dev/r/e8500ee1-9159-4fea-a49b-27a46112101d).  

### Cause and possible solution

I believe I have tracked down the cause of this bug to https://github.com/mglaman/phpstan-drupal/blob/6b4fd0485597d34903efd6a66a6516b659461d2c/src/Type/EntityRepositoryReturnTypeExtension.php#L78-L83.

When `$entityIdArg->getConstantStrings()` is empty (i.e. when no type is statically known) then the union is empty, and collapses to never.
I suspect inserting an extra guard clause with something like:
```php
if (count(entityObjectTypes) === 0) {
  return $returnType;
}
```
would fix the issue.

I intend to test this, and will attempt to make a PR.

tkw1536:
As noted in mglaman/phpstan-drupal#1040, when `EntityRepositoryInterface::getActive` and friends were passed a non-constant string as the `$entity_type_id` parameter, the involved entity type was inferred as *NEVER*.

This PR fixes the issue by introducing a new guard clause, explicitly checking for no guard clauses being involved, and falling back to a generic return type for the function. It also introduces additional tests, checking that the issue does not reoccur.

Fixes mglaman/phpstan-drupal#1040.

### DIFF
--- a/src/Type/EntityRepositoryReturnTypeExtension.php
+++ b/src/Type/EntityRepositoryReturnTypeExtension.php
@@ -77,7 +77,12 @@ final class EntityRepositoryReturnTypeExtension implements DynamicMethodReturnTy
 
         $entityObjectTypes = [];
         $entityIdArg = $scope->getType($methodArgs[0]->value);
-        foreach ($entityIdArg->getConstantStrings() as $constantStringType) {
+
+        $constantStrings = $entityIdArg->getConstantStrings();
+        if (count($constantStrings) === 0) {
+            return $returnType;
+        }
+        foreach ($constantStrings as $constantStringType) {
             $entityObjectTypes[] = $this->entityDataRepository->get($constantStringType->getValue())->getClassType() ?? $returnType;
         }
         $entityTypes = TypeCombinator::union(...$entityObjectTypes);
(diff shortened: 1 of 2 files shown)


tkw1536:
Not sure about the failing check, that seems to also be the case on the main branch.

mglaman:
Thank you!

PR #1055, opened 2026-09-17
Do not report inherited entity storage constructor parameters (#1039)

### LINKED ISSUE #1039: entityStoragePropertyAssignment triggers when extending list builders.
# Bug report

Both EntityListBuilder and DraggableListBuilder (and others?) have EntityStorageInterface as an argument. When you extend those classes, override the constructor and therefore also require a storage argument, the rule triggers. But there's really nothing you can do here as the parent core class requires it.

```
class GroupRoleListBuilder extends DraggableListBuilder {

  public function __construct(EntityTypeInterface $entity_type, EntityStorageInterface $storage, RouteMatchInterface $route_match) {
    parent::__construct($entity_type, $storage);
```

pfrenssen:
EntityListBuilder, ConfigEntityListBuilder and DraggableListBuilder require EntityStorageInterface in their constructors, so a subclass that overrides the constructor to add dependencies has no choice but to accept and forward $storage. EntityStorageDirectInjectionRule now skips as many storage parameters as the nearest inherited constructor declares, resolved through PHPStan reflection rather than a list of core classes. Any further storage parameter is still reported as the subclass's own addition.

EntityStoragePropertyAssignmentRule had the same blind spot for assignments to a property declared by an ancestor, such as EntityListBuilder::$storage, and now only reports properties the class declares itself.

This is authored by Claude Fable under my guidance, I have reviewed and tested the resulting code and went through multiple iterations to improve it. I take responsibility for the resulting code.

Fixes #1039


pfrenssen:
Tests are failing due to #1056

mglaman (commented):
The direct injection side looks good. Resolving the obligation through reflection beats a list of core classes, and the fixtures cover the edge cases. One problem with the property assignment exemption, noted inline. CI failures here are only #1056, fixed on `main` by #1057.

src/Rules/Drupal/EntityStoragePropertyAssignmentRule.php
```diff
@@ -60,4 +67,24 @@ public function processNode(Node $node, Scope $scope): array
...
+        if ($classReflection === null) {
+            return false;
+        }
+
+        $propertyName = $propertyFetch->name->toString();
+        if (!$classReflection->hasInstanceProperty($propertyName)) {
+            return false;
+        }
+
+        $declaringClass = $classReflection->getInstanceProperty($propertyName, $scope)->getDeclaringClass();
```
mglaman, line 98: This exempts a property declared by *any* ancestor, not only one the parent constructor requires. Moving the property into a base class silences the rule completely, since the rule flags assignments, not declarations:

```php
abstract class MyBase {
  protected EntityStorageInterface $nodeStorage;
}

final class MyService extends MyBase {
  public function __construct(EntityTypeManagerInterface $etm) {
    $this->nodeStorage = $etm->getStorage('node'); // not reported
  }
}
```

Declare the same property on `MyService` and it is reported. Narrow the exemption to properties whose declaring class also takes storage in its constructor, the same inherited-obligation check `EntityStorageDirectInjectionRule` uses. `EntityListBuilder::$storage` qualifies, `MyBase::$nodeStorage` does not.

I'll push this change to the branch, along with a fixture case, and merge `main` so CI picks up the #1056 fix.

PR #1057, opened 2026-09-17
Fix CI on main: php-parser ArgPlaceholder and Drupal 11.x fixture

### LINKED ISSUE #1056: Tests workflow fails on main: Linting (php-parser ArgPlaceholder) and Drupal 11.x-dev (RenderCallbackRuleTest)
The scheduled Tests workflow on `main` has been failing on every run since at least 2026-09-14, on two jobs. Both failures are unrelated to any open PR and show up on PR CI as well (e.g. #1055).

Latest failing run on main: https://github.com/mglaman/phpstan-drupal/actions/runs/35176902133

## 1. Linting | PHP 8.3

```
 ------ -----------------------------------------------------------------------
  Line   src/Type/DrupalServiceDynamicReturnTypeExtension.php
 ------ -----------------------------------------------------------------------
  61     Access to an undefined property
         PhpParser\Node\Arg|PhpParser\Node\ArgPlaceholder::$value.
         🪪  property.notFound
 ------ -----------------------------------------------------------------------
```

nikic/php-parser 5.9 added `PhpParser\Node\ArgPlaceholder` (PHP 8.5 partial function application), so `MethodCall::$args` elements are now `Arg|VariadicPlaceholder|ArgPlaceholder`. The code only excludes `VariadicPlaceholder` before reading `->value`:

```php
$arg1 = $methodCall->args[0];
if ($arg1 instanceof VariadicPlaceholder) {
    throw new ShouldNotHappenException();
}

$arg1 = $arg1->value;
```

Replacing the negative check with a positive `instanceof Node\Arg` guard fixes it.

## 2. Tests | PHP 8.3 | Drupal 11.x-dev

`RenderCallbackRuleTest::testRule` (data set #0) fails because Drupal 11.x-dev now produces one extra error for the fixture:

```
--- Expected
+++ Actual
@@ @@
 30: #pre_render callback method 'array{'\\Drupal\\pre_render…', 'unsafeCallback'}' at key '4' does not implement attribute \Drupal\Core\Security\Attribute\TrustedCallback.
     💡 Change record: https://www.drupal.org/node/3349470
+31: #pre_render callback 'user.toolbar_link…' at key '5' is not callable.
```

The expected error list in `tests/src/Rules/RenderCallbackRuleTest.php` needs to account for the `user.toolbar_link…` callback no longer being callable on 11.x-dev. The other Drupal versions in the matrix pass.

pfrenssen:
Guard the first argument of \Drupal::service() with a positive Arg check; nikic/php-parser 5.9 added ArgPlaceholder alongside VariadicPlaceholder.

Drupal 11.x removed the user.toolbar_link_builder service used by the RenderCallbackRule fixture as a trusted service callback. Define the service in the fixture module itself so the test no longer depends on core's service list.

Fixes #1056


src/Type/DrupalServiceDynamicReturnTypeExtension.php
```diff
@@ -54,7 +54,16 @@ public function getTypeFromStaticMethodCall(
...
-        if ($arg1 instanceof VariadicPlaceholder) {
+        // An element of $methodCall->args is one of three node types:
+        // - Arg: a regular argument such as \Drupal::service('foo'). This is
+        //   the only type that carries a value we can inspect.
+        // - VariadicPlaceholder: the "..." in first-class callable syntax,
+        //   \Drupal::service(...).
+        // - ArgPlaceholder (nikic/php-parser 5.9+): the "?" in partial function
+        //   application, \Drupal::service(?).
+        // The two placeholders describe a callable being created, not a call
+        // being made, so there is no service ID to resolve and we bail out.
```
mglaman: The code throws `ShouldNotHappenException` here, so "bail out" reads like an early return. The throw is correct: PHPStan resolves placeholder calls as callables and never sends them to this extension. The comment should say that.

```suggestion
        // The two placeholders create a callable instead of making a call, so
        // PHPStan never asks this extension to resolve them.
```

I'll push this to the branch so we can get CI green.

